Issue #4, opened 2026-04-20
Feature requests: Search function and table resizing


emvolz:
Thanks for making this excellent lightweight md viewer. Two feature requests: 
1.  A simple search and findall function 
2.  The ability to resize the view window in tables (e.g. by click+drag and/or in preferences); on my system, the horizontal scroll does not work unless there is focus on the bottom of the table. Viewing large tables is difficult.

chrismeyersfsu:
Loving using this viewer. This is are my top two feature requests after using it for a week.

PR #5, opened 2026-05-14
Fix wide inline code wrapping

aki1ro:
## Summary
Ran into this odd issue where the left side of my text was disapearing, this occured on wayland, which is what I use. Did some digging with openclaude and found a little workaround. Didnt want to change too much. I'll let you decide if this is an ideal fix for this issue, but works flawless now with any resizing of the window. 

- Use the available pane width for Markdown rendering so wide documents fill the visible content area.
- Split long inline code tokens into bounded chunks so path-like metadata wraps instead of overlapping or clipping.
- Add regression coverage for wide inline code wrapping and unbroken inline code wrapping.
- Verified smooth native Wayland rendering in a wide md-viewer window.

## Test plan
- cargo test --locked
- cargo build --locked
- cargo clippy --locked --all-targets -- -D warnings
- Visual smoke test on a wide native Wayland md-viewer window with long inline metadata paths

## Original Issue:
01-before-overlap-cutoff

## With fix, set the width statically across all window sizes
02-intermediate-wrap-fixed-width-narrow

## Final fix with width filling as far line length will go. 
03-after-full-width-wayland

🤖 Generated with [OpenClaude]([](https://github.com/Gitlawb/openclaude))


aki1ro:
Want me to remove the tests generated? I went ahead and made sure the Agent stayed with your documentation method, so devlog has a new entry which is majority of the line additions.

aydiler:
Thanks for catching this and the detailed screenshots — repro'd locally and confirmed your wrap approach works.

Rather than ping-pong on cleanup, I landed the inline-code wrap fix directly with you credited via `Co-authored-by:` — see 26a356c. A few things I changed before merging:

- Kept `default_width(Some(600))` in `src/main.rs`. The 600 px cap is intentional (55-75 CPL typography, see `docs/LESSONS.md` → "Optimal line length is 55-75 CPL"). I tested the wrap fix at narrow (800 px) and wide (1500 px) widths with the cap kept and the original bug is fully fixed in both. If you'd like full-width rendering as an option, that's worth its own issue/PR — happy to discuss the UX there.
- Dropped `patches/0001-*.patch` (format-patch artifact, not meant to be committed) and `docs/superpowers/specs/2026-05-14-search-find-all-design.md` (the search design doc belongs on #4 — feel free to open a separate PR for that, I'd review it).
- Consolidated the wrap tests into the crate-local location (`crates/egui_commonmark/egui_commonmark/tests/wrapping.rs`) next to the code they exercise. Made the row-height threshold dynamic (queried from ctx) instead of the magic `> 23.0` so the assertions survive line-height changes.
- Added a comment in `inline_code_wrap_segments` noting *why* it's a blind char-count cut rather than break-friendly chars — I tried break-friendly first and it regresses at narrow widths because variable-length segments can still exceed the column.

Closing this in favor of the merged commit. Appreciate the report and the bug fix!

aki1ro:
Sounds good. Will open new PR for toggle enabled width expansion. Also cleanup the repo with superpowers folder added to gitignore as to not dirty up the branches. Will read through your lessons to better understand project trajectory, and continue with testing out the find-all feature. I'll open a PR once it's decent and let you gauge it. 

Thanks!

aydiler:
Hey, heads-up before you sink more time into the find-all PR — I've been actively working on search this week (on `feature/search`). My v1 scope ended up being **current-document only with inline match highlights** (via an extension to the vendored `egui_commonmark` cache: `set_search_ranges`/`set_active_search_range`), `Ctrl+F` / `Enter` / `Shift+Enter` / `Esc`, on-screen `↑`/`↓` buttons + match count, and 9 unit tests on the pure `find_matches`. Cross-tab find-all is deferred to v2.

So to avoid us duplicating effort: I'll land v1 first, then cross-tab find-all is fair game as a separate PR if you'd still like to take it. Up to you — happy to look at the full-width toggle PR in the meantime if you'd rather start there.

aki1ro:
Roger that. Thanks for the heads-up

Issue #22, opened 2026-05-23
Tables scroll slightly horizontally as you scroll past them.


aki1ro:
With the rendering fixed in: https://github.com/aydiler/md-viewer/pull/21. I noticed during scrolling that the tables will catch the scroll wheel action and it "nudges" them either left or right based on whether the user is scroll up past them or down past them. Scrolling up past scrolls the table right, scrolling down scrolls the table left. 

Video evidence below. Not a game breaker at all, just something I noticed. Maybe need a to disable scroll wheel actions within tables? Would need to look at code portions that handle this. 

This issue occurs in default view mode. With full-width enabled (no scroll bars on the tables) the issue is not visible.

https://github.com/user-attachments/assets/ec15a12d-003f-486d-a821-27be436782f4

PR #23, opened 2026-05-24
fix: stop table scroll wheel nudging

aki1ro:
Remove vertical-wheel forwarding from wide table scroll areas so page scrolling no longer shifts tables sideways, while preserving horizontal access through the table scrollbar/native horizontal input. This stemmed from this open issue: https://github.com/aydiler/md-viewer/issues/22

I see it was intended to do that originally, so wondering if we want to keep this behavior? @aydiler I did test this, and scrolling now is smooth without nudging or shifting of the tables left and right. Horizontal bar still works perfectly. Thoughts? 

Video below of before and after. Left is Before, Right is After. 

https://github.com/user-attachments/assets/76ef642f-09fc-4f80-bc79-57b10597fb28

### DIFF
--- a/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
@@ -157,51 +157,6 @@ fn html_cell_visual_lines(cell: &str) -> usize {
     explicit_lines.saturating_add(wrap_est).max(1)
 }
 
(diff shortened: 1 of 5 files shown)


aydiler (approved):
Clean removal — matches the trade-off discussed in #22 (default wheel stays with the page). I'll add a Shift+wheel opt-in in a follow-up so wide-table horizontal scroll is still reachable without dragging the bottom bar.

PR #41, opened 2026-07-01
fix: mermaid text characters render abnormal

MCXCC303:
Added extra parse for 5 double-escaped xml chars
When Mermaid->HTML->SVG, merman will make another escape, but while rendering these chars are only parsed once.

Characters affected:
- `&gt;` -> `>`
- `&lt;` -> `<`
- `&amp;` -> `&`
- `&quot;` -> `"` (potentially affected)
- `&apos;` -> `'` (potentially affected)

Changes has been passed on:
```text
Linux THF-Desktop 6.18.37-1-lts #1 SMP PREEMPT_DYNAMIC Sat, 27 Jun 2026 10:25:22 +0000 x86_64 GNU/Linux
```


MCXCC303:
Before:
screenshot_2026-07-03-191400

After:
screenshot_2026-07-03-191429

PR #42, opened 2026-07-04
Fix bold markdown rendering

### LINKED ISSUE #39: **bold** text not rendered
Hi,

Thanks for md-viewer, it is very nice.

I noticed that **bold** text is not rendered as bold.

aki1ro:
## Summary
- Add an opt-in strong-font path so `**bold**` markdown uses a registered bold family in md-viewer.
- Keep generic `egui_commonmark` consumers safe by leaving the strong font family disabled by default.
- Preserve monospace rendering for strong inline code and document the renderer/font setup in the devlog and lessons.

Fixes aydiler/md-viewer#39

## Test plan
- [x] `cargo test -p egui_commonmark_backend_extended strong --lib` — 3 passed
- [x] `cargo test -p md-viewer` — 29 passed
- [x] `cargo clippy --all-targets --all-features` — passed with pre-existing vendored warnings
- [x] `cargo test` — 29 passed
- [x] `git diff --check`
- [x] Independent verification agent PASS


aydiler (changes requested):
Verified this locally — `**bold**` now renders with genuinely heavier weight, so the fix works and addresses #39. 👍

One issue surfaced during live testing that I'd like resolved before merge:

**Bold text sits on a different baseline and in a different typeface than the surrounding text.** The strong path selects `MarkdownStrong` → **Noto Sans Bold**, but regular body text still renders in egui's bundled **Ubuntu-Light** (Noto Sans is only registered as a *fallback*, appended after the default, so Latin text never reaches it). The two faces have different ascent/x-height metrics, so every `**bold**` span sits a couple px low and looks like a font switch mid-line — visible on a normal sentence and more so in headings/list items.

**Suggested fix:** promote **Noto Sans Regular** from a fallback to the *primary* proportional font in `setup_fonts` (insert at index 0 of the `Proportional` family instead of `push`), so regular and bold both come from the Noto Sans family with matching baselines. CJK/Arabic stay as appended fallbacks. I've verified this locally — it aligns the baseline and unifies the typeface.

**Tradeoff to flag for the maintainer:** this shifts the app's entire body typeface from Ubuntu-Light to Noto Sans (regular text changes appearance too). That's a deliberate, defensible choice for consistent bold, but it's a broader visual change than this PR currently makes — worth an explicit decision.

Everything else in the PR (opt-in default-off for generic consumers, no-panic fallback when the bold font is missing, monospace preserved for strong inline code) looks good.

aydiler:
Thanks for this, @aki1ro — the opt-in strong-font approach is exactly right.

Closing in favor of #51, which includes **all of your commits from this PR** plus a follow-up that fixes the bold baseline. Two reasons for the new branch rather than merging here:

1. This PR now **conflicts with main** — the font/renderer PRs it overlapped (#41, #43, #48, #49) landed first, so it needed rebasing across `main.rs`, `pulldown.rs`, `misc.rs`, and `lib.rs`.
2. On its own it rendered bold in Noto Sans Bold while body text stayed on egui's bundled Ubuntu-Light, so bold sat on a slightly different baseline. #51 promotes Noto Sans to the primary body face so regular + bold share one family and align.

Your work is preserved and credited in #51. Appreciate the contribution! 🙏

PR #75, opened 2026-08-24
fix(math): preserve absolute bars in table formulas

RichardCao:
## Summary

- protect bare vertical bars inside LaTeX-style math pairs before CommonMark table parsing
- decode the protected character before passing the expression to the math backend
- preserve source ranges while preventing formulas such as \(\operatorname{EW}[|\Delta OI|]\) from being split into extra table cells

## Testing

- added a regression test that verifies the table remains at 6 cells and retains the original formula source range
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/egui_commonmark/egui_commonmark/src/parsers/latex_delimiters.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/parsers/latex_delimiters.rs
@@ -29,6 +29,7 @@ use std::ops::Range;
 #[derive(Debug, Clone, PartialEq, Eq)]
 struct Rewrite {
     at: usize,
+    original_len: usize,
     replacement: &'static str,
 }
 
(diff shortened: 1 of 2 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Fixes real breakage: bare `|a|` absolute-value bars inside `\( … \)` pairs were treated by CommonMark's table parser as column separators, splitting cells. The rewrite now masks inner bars as `&#124;` within completed pairs (math backend decodes entities back), with escaped `\|` correctly excluded (`preceding_slashes % 2 == 0`).

Bonus latent-bug fix: `Rewrite.original_len` replaces the hard-coded assumption that every rewrite consumes 2 source bytes — bar rewrites consume 1, so the offset map stays correct for downstream span mapping.

Regression test included. Note for merge order: #82's tests are independent of this fix and pass without it (CI verified on its base).

PR #76, opened 2026-08-24
fix(math): prevent table formulas from clipping

RichardCao:
## Summary

- expand Typst vector frames before rasterization so glyph ink outside the nominal layout bounds is retained
- cache the measured inline-math height
- let table rows reserve the actual formula height and render table math without baseline clipping

This fixes fractions, integrals, bold symbols, and other tall expressions being cut off by table cell boundaries.

## Testing

- math backend tests: 13 passed
- added regressions for vertical ink margins and table row height
- cargo build --locked
- launch smoke test with a Markdown document containing formula-heavy tables
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/egui_commonmark/egui_commonmark/src/lib.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/lib.rs
@@ -84,7 +84,9 @@ pub use egui_commonmark_backend_extended::alerts::{Alert, AlertBundle};
 pub use egui_commonmark_backend_extended::misc::{CommonMarkCache, STRONG_FONT_FAMILY};
 pub use egui_commonmark_backend_extended::typography::{Measurement, TypographyConfig};
 #[cfg(feature = "math")]
-pub use egui_commonmark_backend_extended::render_math;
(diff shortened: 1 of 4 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Root cause of clipped table formulas: baked-in Typst page margins were sized for typical glyphs; fractions/limits/descenders overflow them. The fix moves padding out of Typst and into a post-layout frame transform — translate + grow the vector frame before rasterization. Inline keeps x-padding at zero (preserving the documented tight-spacing rationale for `$w$`/`$z$`) while adding asymmetric y-padding (6pt top / 9pt bottom); display keeps symmetric 8pt/6pt.

Table rows additionally reserve height per cell: exact cached inline-math height when available (new `cached_inline_math_height`), conservative 2× line-height otherwise.

The ink-margin test (`-\frac15\sum…` needs ≥3px top/bottom) directly encodes the bug being fixed.

Branch conflicts with current main in `backend/src/misc.rs` (#92/#83 moved that area) — I can resolve as integrator during landing.

PR #77, opened 2026-08-24
fix(math): add fontconfig CJK fallback

RichardCao:
## Summary

- ask Fontconfig for a system sans-serif font that supports Simplified Chinese
- add the selected face to the Typst math font collection
- keep the existing bundled fonts and only use the system face as a glyph fallback

This allows CJK text inside math expressions to render real glyphs instead of missing-character boxes without hard-coding distribution-specific font names.

## Testing

- added and passed cjk_text_in_math_uses_real_glyphs_when_system_fallback_exists
- verified on Linux with Fontconfig and an installed CJK font
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/egui_commonmark/egui_commonmark_backend/src/misc.rs
+++ b/crates/egui_commonmark/egui_commonmark_backend/src/misc.rs
@@ -390,6 +390,34 @@ mod tests {
         }
     }
 
+    #[cfg(feature = "math")]
+    #[test]
+    fn cjk_text_in_math_uses_real_glyphs_when_system_fallback_exists() {
+        if !MATH_FONTS
+            .iter()
+            .any(|font| font.ttf().glyph_index('在').is_some())
+        {
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

`\text{中文}` inside formulas currently renders as identical tofu boxes because Typst's embedded fonts lack CJK. Fix asks fontconfig for one generic Chinese sans (`fc-match -f '%{file}\n' 'sans-serif:lang=zh-cn'`) and loads just that file into `MATH_FONTS` — fast, portable, no hard-coded Noto/WenQuanYi path.

Degradation is graceful at every step (command failure, empty output, unreadable file → silently skip). Test self-skips when the system lacks CJK coverage, so CI stays green on bare runners.

Note: branch will need a merge against current main — `MATH_FONTS`/test-module region moved with #92/#83. Happy to resolve as integrator if you prefer.

PR #79, opened 2026-08-24
fix(sidebars): allow full labels and flexible widths

RichardCao:
## Summary
- show complete Explorer and Outline labels in horizontal scroll areas
- allow both side panels to resize without an upper width limit
- retain a small splitter target at the minimum width
- persist and restore each panel width across application restarts
- sanitize invalid or obsolete stored width values

## Tests
- cargo test --locked: 48 passed, 1 ignored
- regression coverage for finite, undersized, NaN, and infinite persisted widths


aydiler (approved):
## Review verdict: ✅ Approve — one integration note

Explorer/sidebar labels now render fully (UI elision instead of pre-truncated `display_name` strings), with sane min-widths (70px), wider resize grab targets, and persisted-width sanitization on restore (test included).

**Integration note:** this PR deletes `truncate_display_name`, but #83 (now on main) calls it for outline `display_title`s — textually clean merge, guaranteed compile break. I'll restore that one function during landing resolution so both call sites work; nothing needed from your side.

PR #80, opened 2026-08-24
fix(linux): add Tkinter fallback for file dialogs

RichardCao:
## Summary

- check whether an XDG desktop portal is available before using the native Linux dialog path
- fall back to Python Tkinter when neither the portal path nor the normal native dialog can serve the current desktop environment
- support both file and folder selection
- preserve non-UTF-8 Unix path bytes and surface dialog errors to the user

This helps minimal Linux installations where the application is installed in a user directory and neither XDG Desktop Portal nor Zenity is available. The fallback is conditional and leaves the existing dialog path unchanged on configured desktops.

## Testing

- application tests: 50 passed, 1 ignored
- verified on Linux with Tk 8.6 and without an available desktop portal
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,6 +6,8 @@ use std::fs;
 use std::io::{self, IsTerminal};
 #[cfg(unix)]
 use std::os::unix::net::UnixStream;
+#[cfg(target_os = "linux")]
+use std::os::unix::ffi::OsStringExt;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 use std::sync::mpsc::{self, Receiver};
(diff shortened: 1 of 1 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Pragmatic Linux fallback: when the xdg-desktop-portal FileChooser is unavailable, dialogs fall back to a `python3` + Tkinter helper (embedded script, initial-dir support). Parsing logic is properly factored for testability — portal introspection matching, picker stdout → `PathBuf` (empty output = user cancel, byte-preserving paths), with unit tests covering exactly those pure parts.

One note (non-blocking): the fallback silently assumes `python3` with tkinter may or may not exist — confirmed the chain degrades to the existing error message rather than panicking. Worth documenting in README eventually.

PR #82, opened 2026-08-24
test(math): cover commands before table delimiters

RichardCao:
## Summary

Add a regression for LaTeX-style inline formulas in Markdown tables when the final expression before the closing delimiter contains a command such as epsilon.

The current pre-parse delimiter implementation already handles this correctly. The test preserves both the complete math source and the exact original source range, guarding against the earlier class of one-character truncation bugs without restoring the superseded parser implementation.

## Testing

- targeted regression passed
- git diff --check

### DIFF
--- a/crates/egui_commonmark/egui_commonmark/src/parsers/latex_delimiters.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/parsers/latex_delimiters.rs
@@ -306,6 +306,35 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Pure regression coverage (+29 test lines, no production change): formulas whose content ends in a command right before the closing delimiter (`\(D_t+\epsilon\)`) must survive inside table cells as single inline-math events. Passes independently of #75 (verified green on this branch's base); complements it.

PR #83, opened 2026-08-24
fix(outline): use CommonMark heading source positions

RichardCao:
## Summary

- build Outline entries from pulldown-cmark heading events instead of a line regex
- support formatted headings, inline code, links, Setext headings, and all CommonMark code-block forms
- key rendered heading positions by their original source byte offset
- keep duplicate and emoji-shortcode headings distinct without depending on rendered title text

The Outline parser and renderer now share a source-stable identity, so clicking a formatted or duplicate heading lands on the correct section.

## Testing

- application tests: 48 passed, 1 ignored
- renderer library tests: 37 passed
- added regressions for formatted, linked, Setext, code-block, duplicate, and shortcode headings
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3373,6 +3373,7 @@ dependencies = [
  "mimalloc",
  "notify",
  "notify-debouncer-mini",
+ "pulldown-cmark 0.13.0",
  "regex",
  "rfd",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ serde = { version = "1", features = ["derive"] }
 env_logger = "0.11"
 log = "0.4"
 regex = "1.12.2"
+pulldown-cmark = { version = "0.13", default-features = false }
 fontdb = "0.23"
 sys-locale = "0.3.2"
 ttf-parser = "0.25"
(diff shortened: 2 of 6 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

This kills the outline's two-sources-of-truth problem. On main, the sidebar parsed headings with its own regex while the renderer recorded scroll positions under *normalized-title + occurrence-index* keys — any divergence (setext headings, `#` inside tilde fences, indented/quoted ATX, duplicate titles) silently misdirected outline clicks or dropped entries.

**The redesign:** both sides now key positions by the heading's **source byte offset** via a shared `header_position_key(usize)` exported from the vendored backend; the app's `parse_headers` is rewritten on `pulldown_cmark`'s offset iterator with `Options::all()` — **exactly matching** the options the renderer's own position-recording pass uses (`pulldown.rs:2135`), so the read/write key boundary can't drift. Duplicate-title fragility disappears by construction: byte offsets are unique.

Verified:
- Setext headings now appear in the outline and jump correctly
- `# Hidden heading` inside `~~~` fences excluded
- Titles compose Text/Code/math/footnote-reference events with sensible separators; formatting stripped (`**Bold**` → "Bold")
- Line numbers computed via `partition_point` over line starts — O(log n) per heading
- The app's new `pulldown-cmark 0.13` dependency is the same crate/version already vendored — no new dependency in practice

Tests updated plus a new formatting/setext/fenced-heading case. CI green.

Nit (non-blocking): both sides spell `Options::all()` literally today; deriving them from one shared constant would make future option changes structurally safe.

PR #84, opened 2026-08-24
perf(explorer): scan roots asynchronously without losing expansion

RichardCao:
## Summary
- restore expanded directory children when an asynchronous Explorer result is applied
- move both local and GVFS root scans/refreshes off the UI thread
- centralize scan startup and fall back to synchronous scanning if thread creation fails
- ignore stale results whose root no longer matches
- reapply the current sort and expansion state when results arrive

This extends the existing GVFS-only asynchronous path instead of adding a second competing scan mechanism.

## Tests
- asynchronous results restore directories expanded after the scan started
- local root scans complete asynchronously and restore expanded children

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -314,7 +314,7 @@ struct FileExplorer {
     tree: Vec<FileTreeNode>,
     expanded_dirs: HashSet<PathBuf>,
     sort_order: SortOrder,
-    /// Receiver for async directory scan results (GVFS paths scan in background)
+    /// Receiver for asynchronous root-directory scan results.
     pending_scan: Option<Receiver<ExplorerScanResult>>,
 }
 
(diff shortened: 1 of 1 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Generalizes the async root scan from GVFS-only to **all roots** — large local/network directories no longer block the UI on `set_root`/`refresh`.

**Correctness verified:**
- Expansion survives the async boundary twice over: the worker gets a snapshot of `expanded_dirs` (restored depth-sorted so parents load before lazy descendants), and `poll_pending_scan` re-restores from *current* state on apply — covering directories expanded while the worker ran.
- Fixes a race on main: the old poll applied any pending result unconditionally; now a result whose root no longer matches is discarded, and each `set_root`/`refresh` supersedes the previous receiver entirely.
- Sort-order changes while a scan is in flight are handled by re-sorting the result on apply.
- Both `.expect("failed to spawn …")` panics are gone — spawn failure falls back to a synchronous scan.

**Tests** cover async restore of expanded children and local-root asynchronous scanning. CI green.

Interplay with #87 is complementary (#84 = full scans async, #87 = targeted sync refresh) and pairwise conflict-free per merge-tree analysis.

PR #85, opened 2026-08-24
fix(watcher): bound recovery attempts reliably

RichardCao:
## Summary

- separate user-requested watcher starts from automatic recovery attempts
- avoid resetting the retry counter merely because watcher construction succeeded
- stop recovery at the configured limit and surface a terminal error
- report watcher bridge thread startup failures instead of panicking

This prevents repeated disconnects before the first successful event from restarting forever.

## Testing

- application tests: 48 passed, 1 ignored
- added a boundary regression for the complete retry sequence
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -41,6 +41,10 @@ static LINK_RE: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"\[([^\]]*)\]\(([
 const MAX_WATCHER_RETRIES: u32 = 3;
 const FLASH_DURATION_MS: u64 = 600;
 
+fn next_watcher_retry(current: u32) -> Option<u32> {
+    (current < MAX_WATCHER_RETRIES).then(|| current + 1)
+}
+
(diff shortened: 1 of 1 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Fixes a genuine infinite-retry loop on main: `start_watching()` reset `watcher_retry_count = 0` on every call, and *both* recovery sites called it after incrementing — so the counter never accumulated past 0 and a persistently failing watcher retried every ~2 s forever.

The split into `start_watching()` (user action — resets the counter) vs `start_watching_attempt()` (recovery — does not reset) is exactly the right shape. `next_watcher_retry` caps attempts at 3, is overflow-safe (`u32::MAX` covered by test), and exhaustion sets a visible error message plus disables watching instead of hammering.

Bonus robustness: the watcher bridge thread spawn `.expect(…)` panic is replaced with graceful error handling.

Test covers the full retry sequence including the cap.

PR #87, opened 2026-08-24
perf: refresh only changed explorer directories

RichardCao:
## Summary
- rescan only the parent directories affected by watcher events
- preserve unrelated loaded Explorer subtrees and expansion state
- restore expanded descendants in parent-before-child order after a targeted rescan

## Tests
- added a regression test proving an updated directory is refreshed while an unrelated expanded subtree remains loaded
- cargo test --locked: 48 passed, 1 ignored

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -470,6 +470,53 @@ impl FileExplorer {
         }
     }
 
+    /// Rescan only one changed directory while preserving the rest of the tree.
+    fn refresh_directory(&mut self, directory: &Path) {
+        let Some(root) = self.root.as_ref() else {
+            return;
+        };
+        if directory == root {
+            self.tree = Self::scan_directory_shallow(&root.clone(), self.sort_order);
+        } else {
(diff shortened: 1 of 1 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Nice perf win: watcher events previously triggered a full tree rebuild (`refresh()`); now only the parent directory of each changed path is rescanned and spliced into the tree via recursive `replace_directory_children`.

Edge cases check out:
- Change at the root → full shallow rescan (equivalent to old behavior), then shared expansion-restore loop.
- Changed directory not present in the tree (collapsed/lazy) → no-op; fresh content loads on expand. Old code produced the same visible result with much more work.
- Nested expanded dirs under a replaced node are reloaded by the restore loop (depth-sorted).

The regression test verifies unrelated subtrees keep their contents *and* expansion state after a targeted refresh.

One nit (non-blocking): the restore loop iterates *all* expanded dirs after every targeted refresh rather than just those under the replaced directory — bounded by expansion count, fine as-is.

PR #88, opened 2026-08-24
fix: bound MIME SVG caching and defer font loading

RichardCao:
## Summary
- cap cached MIME-detected SVG rasters at 64 MiB
- evict least-recently-used entries when the budget is exceeded
- avoid retaining a single raster larger than the cache budget while still rendering it
- initialize SVG options and scan system fonts only when the first MIME SVG is rendered
- keep forget, forget_all, and byte_size behavior intact

The lazy initialization belongs in this PR because both changes govern the lifetime and resource cost of the same MIME SVG loader. Keeping it here avoids a dependent follow-up PR and uses standard-library OnceLock without another synchronization layer.

## Tests
- added LRU eviction and lazy-initialization regression tests
- cargo +1.89.0 test --manifest-path crates/egui_commonmark/Cargo.toml -p egui_commonmark_extended --features svg,svg_text mime_svg_loader (5 passed)

### DIFF
--- a/crates/egui_commonmark/egui_commonmark/src/mime_svg_loader.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/mime_svg_loader.rs
@@ -4,7 +4,11 @@
 //! loader currently accepts only URIs ending in `.svg`. Dynamic image URLs
 //! commonly omit that suffix while returning `image/svg+xml` correctly.
 
-use std::{collections::HashMap, mem::size_of, sync::Arc};
+use std::{
+    collections::HashMap,
+    mem::size_of,
+    sync::{Arc, OnceLock},
+};
 
 use egui::{
     load::{BytesPoll, ImageLoadResult, ImageLoader, ImagePoll, LoadError, SizeHint},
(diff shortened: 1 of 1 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Full diff reviewed; **tests executed locally** (renderer CI job from #91 not on this branch yet):

```
mime_svg_loader: 5 passed, exit 0
  incl. cache_limit_evicts_least_recently_used_entries
        svg_options_are_initialized_lazily
```

**Verified:**
- LRU stamps `last_used` on both hit and insert with a monotonic tick; a freshly inserted entry always carries the newest tick, so insert-evict thrash of new rasters is impossible.
- Oversized (>64 MiB) rasters are never cached — rendered for the current frame, returned, re-rendered on demand. Matches the stated "render but don't retain" contract; correct tradeoff at that size.
- `OnceLock` defers `load_system_fonts()` + `fc-match` to the first actual MIME-SVG render — strictly less startup work for users who never show one. All accessors (`forget`, `forget_all`, `byte_size`) correctly adapted.

**Non-blocking nits:**
1. `trim_svg_cache` recomputes the full byte sum on every eviction loop (O(n²)) while holding the loader lock on the UI thread. Fine at realistic badge counts; #89 implements the O(n) running-total variant — worth aligning on that pattern here too.
2. The deferral moves the font-scan hitch rather than removing it: a document whose *first* image is a MIME-SVG now pays the scan mid-frame. Acceptable tradeoff; just where I'd look if first-SVG jank is ever reported.

PR #89, opened 2026-08-24
fix: bound embedded data URL resources

RichardCao:
## Summary
- reject individual encoded data URLs larger than 16 MiB before parsing
- bound the decoded data URL cache to 64 MiB with LRU eviction, including URI-key memory
- replace one-thread-per-URL decoding with two reusable workers and a bounded four-job queue
- retry cleanly when the queue is full and avoid restoring entries forgotten while decoding

The changes are split into three focused commits: per-item input limit, total cache budget, and decode concurrency.

## Tests
- strict encoded-size boundary
- normalized `data:` scheme recognition
- cache accounting and LRU eviction
- successful decode-job cache transition
- forgotten pending jobs remain absent

### DIFF
--- a/crates/egui_commonmark/egui_commonmark_backend/src/data_url_loader.rs
+++ b/crates/egui_commonmark/egui_commonmark_backend/src/data_url_loader.rs
@@ -19,6 +19,26 @@ struct Data {
 
 type Entry = Poll<Result<Data, String>>;
 
+const MAX_DATA_URL_ENCODED_BYTES: usize = 16 * 1024 * 1024;
+
+fn data_url_too_large(uri: &str) -> bool {
+    uri.len() > MAX_DATA_URL_ENCODED_BYTES
+}
+
+/// Match the scheme normalization used by `data_url::DataUrl::process`
+/// without scanning the potentially large header or body.
+fn has_data_url_scheme(uri: &str) -> bool {
+    let mut bytes = uri
+        .trim_start_matches(|ch| ch <= ' ')
+        .bytes()
+        .filter(|byte| !matches!(byte, b'\t' | b'\n' | b'\r'));
+    b"data:".iter().all(|expected| {
+        bytes
+            .next()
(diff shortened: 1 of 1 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Full diff reviewed; **tests executed locally** (renderer CI job from #91 not on this branch yet):

```
data_url_loader: 5 passed, exit 0
  incl. cache_budget_includes_uri_keys_and_evicts_lru_entries
        forgotten_pending_job_does_not_restore_cache_entry
```

**Verified layer by layer:**
- Three-stage funnel in `load()`: cheap ASCII scheme check → **16 MiB encoded-size rejection** → expensive full parse only afterwards. Oversized URIs never reach the parser and get a clear error message instead of silent OOM. Scheme pre-check handles leading whitespace / embedded `\t\n\r` per `data_url`'s normalization, with `"database:value"` lookalike tests.
- Budget accounting includes URI-key memory (`uri.len() + payload`) — which most LRU implementations forget — and uses a running total with `saturating_sub`, so trim is O(n) rather than per-eviction resummation.
- Pool/backpressure: same proven shared-receiver pattern as #92 (2 workers, `sync_channel(4)`). On queue-full: pending marker removed + `request_repaint_after(50ms)` backoff. On disconnected channel: visible error state covering zero-worker failure.
- Race protection: `decode_job` writes only over still-pending entries, so a `forget()` between enqueue and completion cannot resurrect a stale result; eviction deliberately skips pending entries; dedicated regression test.

**Non-blocking nits:**
1. Each queued `DecodeJob` owns a full clone of its URI string (≤16 MiB) while an equal copy sits in the cache key — pathological all-max-size input transiently reaches ~2× the advertised budget. Sharing an `Arc<str>` between key and job would halve peak usage.
2. Hard-coded 2 workers / bound-4 vs #92's policy-driven sizing — fine independently; consolidation opportunity if both land.

PR #91, opened 2026-08-24
ci: cover the vendored renderer workspace

RichardCao:
## Summary
- add a dedicated CI job for the nested `crates/egui_commonmark` workspace
- lint and test the renderer, backend, and macro crates with the feature sets used by md-viewer
- repair stale example and doctest imports left behind by the `*_extended` crate rename
- update macro code generation for the current code-block API and make `commonmark_str!` paths robust in trybuild/Cargo contexts

The root Cargo commands compile the path-patched renderer as a dependency, but did not run its own unit, integration, doctest, example, or proc-macro tests. Enabling that coverage exposed the stale targets fixed in the first commit.

Existing renderer warnings are reported but are not promoted to errors in this PR; warning cleanup can be reviewed separately.

## Verification
- Clippy for all three nested crates and renderer examples
- 40 renderer unit tests, 12 wrapping tests, and 16 doctests
- 14 backend unit tests and its doctest
- proc-macro trybuild suite and doctest

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,44 @@ jobs:
       - run: cargo build --release
       - run: cargo test
 
+  renderer:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: read
+    env:
+      RENDERER_FEATURES: better_syntax_highlighting,svg,svg_text,load-images,fetch,mermaid,math,macros
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang pkg-config libxcb1-dev libxcb-render0-dev \
+            libxcb-shape0-dev libxcb-xfixes0-dev libxkbcommon-dev libssl-dev \
+            libgtk-3-dev libfontconfig1-dev libdbus-1-dev
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: crates/egui_commonmark -> target
+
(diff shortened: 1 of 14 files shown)


aydiler (approved):
## Review verdict: ✅ Approve — merge first

Reviewed commit-by-commit against `main` (121ae04). Clean split: `f07e142` = repairs, `3ff9a4f` = workflow.

**Why this is the keystone:** it fixes real breakage on main today — generated code in `egui_commonmark_macros/src/generator.rs` still calls `CodeBlock::end(ui, cache, &options, max_width)` with 4 args while `backend/src/misc.rs:547` already requires 5 (`id: egui::Id`). Anyone using `commonmark!`/`commonmark_str!` with a code block gets broken codegen right now. This PR is the proof of concept for its own necessity.

**Verified:**
- `RENDERER_FEATURES` exactly matches the root dependency feature list + `macros` (justified: examples/tests exercise the macro path)
- Backend linted/tested with `--all-features`, dedicated rust-cache workspace scope
- All example/doctest diffs are pure mechanical `egui_commonmark::*` → `egui_commonmark_extended::*` renames
- `"tests/file.md"` path fix pairs correctly with the new `CARGO_MANIFEST_DIR` candidate; fixture exists; `tests/fail/*` fall-through preserved

**Two non-blocking nits (could be fast follow-ups):**
1. `macros/src/lib.rs`: the third path candidate uses `env!("CARGO_MANIFEST_DIR")`, which expands at proc-macro-*compile* time and bakes the vendored crate's build dir into the dylib — at downstream invocation it resolves against the wrong crate. Candidate 2 (`std::env::var_os`) is the only correct lookup; suggest deleting candidate 3.
2. Resolution order tries cwd-relative first; cargo runs rustc from workspace root, so a same-named file at repo root silently shadows one next to the invoking crate. Worth a doc comment.

Also unblocks #93 (explicitly staged behind this).

PR #92, opened 2026-08-24
fix(math): bound rendering with a shared worker pool

RichardCao:
## Summary
- replace one OS thread per formula with one process-wide worker pool
- size workers from the existing bounded `math_concurrency()` policy
- bound queued formula jobs to 64 and retry naturally on later repaint when full
- surface a stable error if worker startup becomes unavailable
- include exact foreground and background colors in math texture cache keys

The existing per-document active-render limit remains in place. The shared pool adds a process-wide bound across multiple tabs/caches without introducing the old branch's separate global raster cache.

## Verification
- all 16 backend tests with the `math` feature
- shared-worker end-to-end render test
- cache-key color regression test
- root `cargo check --locked`

### DIFF
--- a/crates/egui_commonmark/egui_commonmark_backend/src/misc.rs
+++ b/crates/egui_commonmark/egui_commonmark_backend/src/misc.rs
@@ -4,13 +4,19 @@ use egui::{RichText, TextStyle, Ui, text::LayoutJob};
 use std::collections::HashMap;
 #[cfg(feature = "math")]
 use std::collections::HashSet;
-#[cfg(any(feature = "better_syntax_highlighting", feature = "mermaid"))]
+#[cfg(any(
+    feature = "better_syntax_highlighting",
+    feature = "mermaid",
+    feature = "math"
+))]
 use std::sync::Arc;
 
 use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
 #[cfg(any(feature = "mermaid", feature = "math"))]
 use std::sync::mpsc;
+#[cfg(feature = "math")]
+use std::sync::Mutex as StdMutex;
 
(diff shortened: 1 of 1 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Full diff reviewed; **tests also executed locally** (this branch predates the renderer CI job from #91, so its vendored-crate tests don't run in CI yet):

```
test misc::tests::math_cache_key_includes_exact_colors ... ok
test misc::tests::shared_math_worker_returns_rendered_formula ... ok
3 passed; exit 0
```

**Design verified sound:**
- Global `LazyLock` pool sized by the existing `math_concurrency()` policy (1–6); shared-receiver pattern locks the `Mutex<Receiver>` only for pickup and renders outside the lock, so real parallelism is preserved.
- Backpressure is airtight: `try_send`; on `Full` the hash stays unmarked so the existing spawn gate (`!contains && len < math_concurrency()`) retries naturally next frame. The gate caps in-flight markers at ≤6, so the 64-slot queue is defensive headroom — no unbounded growth.
- Zero-worker edge case covered: if all spawns fail, receivers drop → channel disconnects → visible `MathState::Error("math renderer workers are unavailable")` instead of an eternal placeholder.
- Bonus correctness fix: cache key now hashes exact fg/bg colors instead of the `dark_mode` bool — theme tweaks that don't flip dark/light invalidate correctly (dedicated test included).

**Non-blocking nits:**
1. `Full` path is silent — a `log::debug!` would help diagnose future "formula appears late" reports.
2. Test uses the global pool; a second pool-touching test in the same binary could make queue-fullness order-dependent. Fine today.

Merge order note: land #91 first, then update this branch so the new `renderer` check runs on it.

Issue #95, opened 2026-08-24
Possible fix(deps): rustls-webpki 0.103.9 → 0.103.13, 0.104.0-alpha.7 (GHSA-82j2-j2ch-gfr8) in Cargo.lock


begininvoke:
Came across something in `crates/egui_commonmark/Cargo.lock` around line 4181 that looked worth flagging.

The `bit_string_flags()` function in `rustls-webpki` mishandles a BIT STRING whose content is exactly `[0x00]` (zero padding bits and no data bytes). The guard clause allows `raw_bits` to be empty when `padding_bits == 0`, then unconditionally indexes `raw_bits[raw_bits.len() - 1]`, causing an under‑flow and out‑of‑bounds panic. This panic can be triggered via the public `BorrowedCertRevocationList::from_der()` API when a crafted CRL with an `issuingDistributionPoint` extension containing such a BIT STRING is supplied. Applications that enable CRL revocation checking and accept attacker‑controlled CRL data can be crashed, leading to a denial‑of‑service. The vulnerability is classified **HIGH** because it can be exploited without authentication and results in immediate process termination.

**Impact:** Any rustls‑webpki user that opts‑in to CRL verification (via `RevocationOptions`) and loads untrusted CRL bytes is vulnerable. An attacker can supply a malicious CRL (e.g., via a compromised CDP URL) to cause the TLS handshake to panic, leading to a DoS. The default configuration (no CRL checking) is not affected.

**Risk level:** High – easy to trigger, no privileges required, results in service outage.

Something like this might fix it:

```
*** Begin Patch
*** Update File: src/der.rs
@@
-    if padding_bits > 7 || (raw_bits.is_empty() && padding_bits != 0) {
-        return Err(Error::BadDer);
-    }
-    let last_byte = raw_bits[raw_bits.len() - 1];  // ← crash
+    // NOTE: A BIT STRING with a single padding‑bits byte of value 0 (i.e.
+    // `padding_bits == 0` and `raw_bits` is empty) is a valid encoding that
+    // represents an empty set of flags. The original check allowed this case
+    // to pass, but the subsequent indexing of `raw_bits` caused an under‑flow
+    // and panic. We now treat the empty‑bits case as a valid, all‑zero value.
+    if padding_bits > 7 || (raw_bits.is_empty() && padding_bits != 0) {
+        return Err(Error::BadDer);
+    }
+
+    // Safely obtain the last byte only when there is at least one data byte.
[...]

PR #99, opened 2026-08-25
fix(deps): update nested rustls-webpki lock

RichardCao:
Addresses #95.

The root lockfile already resolves `rustls-webpki` to 0.103.15, but the standalone `crates/egui_commonmark/Cargo.lock` still pinned 0.103.9. This updates that nested lockfile to 0.103.15, above the 0.103.13 patched threshold for GHSA-82j2-j2ch-gfr8 / RUSTSEC-2026-0104.

No source dependency constraint changes are needed: the existing semver range already accepts the patched release.

Validation: `cargo check --manifest-path crates/egui_commonmark/Cargo.toml --locked`.

### DIFF
--- a/crates/egui_commonmark/Cargo.lock
+++ b/crates/egui_commonmark/Cargo.lock
@@ -4180,9 +4180,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.9"
+version = "0.103.15"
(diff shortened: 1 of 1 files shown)


RichardCao:
Upstream PR #97 was merged while this PR was being prepared and applies the same vendored-lockfile update to rustls-webpki 0.103.15. Closing this duplicate; no additional source constraint change is needed.

PR #100, opened 2026-08-25
perf(svg): trim raster cache in one eviction pass

RichardCao:
Follow-up to #88.

The SVG cache trim loop recomputed the full byte total and rescanned the map for the LRU entry for every eviction. A large over-budget trim was therefore quadratic in the number of cached entries.

This change computes the total once, builds and sorts the LRU eviction order once, and then removes entries until the byte limit is met. It also adds regression coverage for evicting multiple entries in the correct order.

Validation:
- `cargo test --manifest-path crates/egui_commonmark/Cargo.toml -p egui_commonmark_extended --features svg cache_limit`
- `cargo check --locked`
- `git diff --check`


aydiler (approved):
## Review verdict: ✅ Approve

Implements the O(n) eviction suggested during #88's review. Victim list is built once, sorted by `last_used`, evicted in order with a running total (`saturating_sub`) — no more full-map rescan + sum per victim (was O(n²) under the loader lock on the UI thread). Regression test included.

PR #101, opened 2026-08-25
perf(loader): share data URL strings with decode jobs

RichardCao:
Follow-up to #89.

Encoded data URLs can be as large as 16 MiB. The pending cache key and queued decode job previously each owned a separate `String`, temporarily doubling that encoded allocation before decoding.

This changes cache keys and decode jobs to `Arc<str>` so both owners share one allocation. Borrowed `&str` lookup/removal remains available, and the LRU byte budget continues to include the URI length. A regression test verifies that the pending cache entry and job point to the same allocation.

Validation:
- `cargo test --manifest-path crates/egui_commonmark/Cargo.toml -p egui_commonmark_backend_extended --features embedded_image,math data_url_loader` (6 passed)
- `cargo check --locked`
- `git diff --check`

### DIFF
--- a/crates/egui_commonmark/egui_commonmark_backend/src/data_url_loader.rs
+++ b/crates/egui_commonmark/egui_commonmark_backend/src/data_url_loader.rs
@@ -26,13 +26,13 @@ struct CacheEntry {
 
 #[derive(Default)]
 struct DataUrlCache {
-    entries: HashMap<String, CacheEntry>,
+    entries: HashMap<Arc<str>, CacheEntry>,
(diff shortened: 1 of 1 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Implements the `Arc<str>` sharing suggested during #89's review: one allocation shared between the pending cache entry and the queued `DecodeJob`, halving worst-case transient memory for max-size URIs. `trim_cache`/lookups adapted via `as_str()`; a dedicated test proves cache-entry and decode-job share the allocation.

PR #102, opened 2026-08-25
fix(macros): resolve markdown from invoking manifest

RichardCao:
Follow-up to the renderer-workspace integration in #91.

`commonmark_str!` previously tried rustc's raw working directory before the invoking crate manifest and also baked the proc-macro crate's own compile-time manifest path into every build. That can select the wrong same-named file or accidentally read a file from the dependency source tree.

This change:
- resolves relative paths against the invoking crate's `CARGO_MANIFEST_DIR` first
- retains the runtime cwd as a compatibility fallback
- treats absolute paths directly and deduplicates identical candidates
- removes the proc-macro crate's baked manifest fallback
- tracks the resolved path for the nightly tracking feature
- adds unit coverage for precedence, absolute paths, and deduplication

Validation:
- `cargo test --manifest-path crates/egui_commonmark/Cargo.toml -p egui_commonmark_macros_extended`
- `cargo check --locked`
- `git diff --check`

### DIFF
--- a/crates/egui_commonmark/egui_commonmark_macros/src/lib.rs
+++ b/crates/egui_commonmark/egui_commonmark_macros/src/lib.rs
@@ -77,6 +77,7 @@ mod generator;
 use generator::*;
 
 use quote::quote_spanned;
+use std::path::{Path, PathBuf};
 use syn::parse::{Parse, ParseStream, Result};
 use syn::{Expr, LitStr, Token, parse_macro_input};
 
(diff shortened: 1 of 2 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Resolves both nits from #91's review, properly:
- The compile-time `env!("CARGO_MANIFEST_DIR")` candidate is gone; resolution is now a pure, unit-tested `markdown_path_candidates()` (absolute short-circuit; manifest-dir preferred, cwd fallback deduped)
- Bonus: `tracked_path` now records the *resolved* path, so change detection no longer depends on rustc's cwd

PR #103, opened 2026-08-25
fix(math): retry saturated render queues

RichardCao:
## Summary
- report bounded math queue saturation instead of silently dropping the enqueue attempt
- request a prompt repaint so saturated jobs are retried as workers drain the queue
- construct an isolated worker pool in tests instead of competing with the process-global pool
- restore missing test attributes on the CJK fallback and inline fraction raster regressions

## Validation
- backend tests with math enabled: 19 passed
- backend tests with default features: 13 passed
- root locked cargo check passed


aydiler (approved):
## Review verdict: ✅ Approve

Refactors the static `LazyLock` pool into a parameterized `start_math_workers(worker_count, capacity)` + tri-state `enqueue_math_job()` (`Queued`/`Full`/`Disconnected`) — making the backpressure paths from #92 unit-testable for the first time. Tests cover queue-full reporting and end-to-end formula delivery.

PR #104, opened 2026-08-25
chore: clear actionable compiler warnings

RichardCao:
## Summary
- make f32 stroke widths explicit for compatibility with the new future-incompatible lint
- mark the retained public code-block width parameter as intentionally unused
- discard unused table scope responses explicitly
- replace the deprecated heading child-UI helper with its exact scope_builder equivalent
- remove the unused parser_options_math helper left behind after parser integration

The parser helper is not referenced by #96 or the current renderer. Removing it is independent of viewport clipping and prevents the dead-code warning from returning after the rest of this cleanup.

## Validation
- root locked cargo check passed without source warnings
- egui_commonmark library tests with math enabled: 43 passed

Default-feature workspace tests currently hit the pre-existing missing math test guards fixed separately in #103; all-features also enables the nightly-only feature on stable.

### DIFF
--- a/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
@@ -437,14 +437,6 @@ impl CommonMarkViewerInternal {
     }
 }
 
-fn parser_options_math(is_math_enabled: bool) -> pulldown_cmark::Options {
-    if is_math_enabled {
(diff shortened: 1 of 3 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Clears the actionable vendored-crate warnings we kept seeing in build logs: explicit `_f32` float literal suffixes (future-incompatible fallback), `_max_width`, and friends. Mechanical, safe.

PR #105, opened 2026-08-25
docs(linux): explain file dialog dependencies

RichardCao:
## Summary
- document the two Linux file-dialog paths: XDG Desktop Portal with gdbus, or Python 3 with Tkinter
- list common Tkinter package names for Debian/Ubuntu, Fedora/RHEL, and Arch
- include the Arch fallback packages in the existing dependency command
- correct source comments that previously described Tkinter as no-install

## Validation
- verified the documented Tkinter import command locally


aydiler (approved):
## Review verdict: ✅ Approve

Docs follow-up to #80's Tkinter dialog fallback — README now states the portal/Tkinter requirements instead of letting Linux users hit a silent failure. Correct pairing.

PR #107, opened 2026-08-25
fix(fonts): use platform script fallbacks

### LINKED ISSUE #106: Unicode issues on Mac
On my Mac, Unicode - that is to say, non-Latin scripts and symbols - render as tofu. I asked Claude to look into it, and it said the following. Hope this is helpful.

---

**Title:** Font fallback misses installed scripts on macOS (Thai, Devanagari, Simplified Chinese)

## Summary

`setup_fonts` picks fallback fonts by hardcoded family *name*. The list in
`REGULAR_FONT_SPECS` (`src/system_fonts.rs`) is Linux/Windows-oriented, so on macOS
most specs resolve to nothing and several scripts render as tofu boxes — even
though the system has perfectly good fonts for them.

## Repro

macOS 15 (Darwin 25.6.0), md-viewer @ 886dc93, `cargo build` + open this file:

```
Em dash: A — B. Ellipsis: a…b
Thai: สวัสดีครับ ภาษาไทย
Devanagari: नमस्ते हिन्दी
Chinese: 中文测试   Japanese: こんにちは   Korean: 한국어
````

Result: Thai and Devanagari render as boxes. 中文 renders but 测试 is boxes.
Everything else (em/en dash, ellipsis, curly quotes, arrows, symbols, math, Greek,
Cyrillic, Arabic, Hebrew, JP, KR) renders fine.

The same Thai text renders correctly in vim in the same terminal, so the fonts are
installed — md-viewer just never asks for them.

Root cause

I probed all 853 faces fontdb::load_system_fonts() finds on this machine against
the app's own spec list. Only 3 of 12 specs resolve:
┌─────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────┐
│                      Spec                       │                                      macOS result                                      │
├─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤
│ SystemSans                                      │ ✅ Arial — but only on the 10th candidate; the 9 before it are Linux/Windows names     │
├─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤
│ SystemCjkJpSans                                 │ ✅ Hiragino Sans                                                                       │
├─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤
│ SystemCjkKrSans                                 │ ✅ Apple SD Gothic Neo                                                                 │
├─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤
│ SystemThaiSans                                  │ ❌ asks only for Noto Sans Thai                                                        │
├─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤
│ SystemDevanagariSans                            │ ❌ asks only for Noto Sans Devanagari                                                  │
├─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤
│ SystemCjkScSans                                 │ ❌ all 10 names absent (incl. PingFang SC, which no longer exists at that family name) │
├─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤
│ SystemSymbols, SystemSymbols2, SystemDejaVuSans │ ❌ absent                                                                              │
├─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────┤
│ SystemArabicSans, SystemHebrewSans              │ ❌ absent (these render only because Arial happens to cover them)                      │
└─────────────────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────┘

**Suggested fix**

Prefer coverage-driven selection over name-driven. The code already parses every
face with ttf_parser, so instead of guessing names per OS, enumerate db.faces()
once and select per script by actual glyph coverage. I measured this locally: ~29 ms
for 15 scripts across 853 faces, on top of the ~34 ms load_system_fonts() already
costs — one-time at startup. Keeping preferred names within a script (for
quality/regional variants) and falling back to coverage only when none resolve gets
both correctness and control.

This also stops the per-OS-name treadmill — docs/LESSONS.md already records the same
class of gap being patched for Windows CJK ("Windows CJK fallback needs Windows font
paths"); macOS is the next instance of it, and the next platform will be another.

Image

RichardCao:
Closes #106.

## Summary
- replace the hardcoded installed-family name lists with Fontique platform fallback discovery
- ask CoreText, fontconfig, and DirectWrite for generic and ISO 15924 script fallbacks using the system locale
- verify actual glyph coverage and search all faces within a returned family, including families that split Latin, Thai, or Devanagari into separate faces
- preserve font-collection face indices and pair true bold faces with each selected regular fallback
- retain egui bundled fonts when the platform cannot provide a suitable face

No installed font family names are hardcoded. The remaining script identifiers and short text samples describe required coverage rather than a distribution or OS font package.

The current workspace MSRV is Rust 1.89; Fontique 0.11.1 requires Rust 1.88.

## Validation
- `cargo check --all-targets`
- system-font unit tests: 3 passed
- ignored installed-font integration test passed locally on Linux for simplified/traditional Chinese, Japanese, Korean, Devanagari, and Thai in both regular and strong font chains
- macOS compilation could not be completed from the Linux host because an Apple C toolchain/sysroot is unavailable; the CoreText path therefore still needs CI or reporter validation on macOS

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,7 +94,7 @@ dependencies = [
  "accesskit_consumer",
  "hashbrown 0.15.5",
  "static_assertions",
- "windows",
+ "windows 0.61.3",
  "windows-core 0.61.2",
 ]
 
@@ -1871,12 +1871,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "font-types"
+version = "0.12.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e64eb721ca85a34323425f4041adc5d82704d3782d5f8f03793bc012419dce23"
+dependencies = [
+ "bytemuck",
+]
+
+[[package]]
 name = "fontconfig-parser"
 version = "0.5.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bbc773e24e02d4ddd8395fd30dc147524273a83e54e0f312d986ea30de5f5646"
 dependencies = [
- "roxmltree",
+ "roxmltree 0.20.0",
 ]
 
 [[package]]
@@ -1891,6 +1900,28 @@ dependencies = [
  "slotmap",
  "tinyvec",
  "ttf-parser",
+]
+
+[[package]]
+name = "fontique"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6688bc1294fe7117d788937b6c53480169b29c566954af490830d4c09da9516a"
+dependencies = [
+ "hashbrown 0.17.1",
+ "linebender_resource_handle",
+ "memmap2",
+ "objc2 0.6.4",
+ "objc2-core-foundation",
+ "objc2-core-text",
+ "objc2-foundation 0.3.2",
+ "parlance",
+ "read-fonts 0.41.0",
+ "roxmltree 0.21.1",
+ "smallvec",
+ "windows 0.62.2",
+ "windows-core 0.62.2",
+ "yeslogic-fontconfig-sys",
 ]
 
 [[package]]
@@ -3136,6 +3167,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "linebender_resource_handle"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d4a5ff6bcca6c4867b1c4fd4ef63e4db7436ef363e0ad7531d1558856bae64f4"
+
+[[package]]
 name = "linked-hash-map"
 version = "0.5.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3314,7 +3351,7 @@ dependencies = [
  "egui",
  "egui_commonmark_extended",
  "env_logger",
- "fontdb",
+ "fontique",
  "image",
  "log",
  "mimalloc",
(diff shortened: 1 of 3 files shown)


aydiler (approved):
## Review verdict: ✅ Approve

Targets #106 (macOS non-Latin scripts rendering as tofu): replaces hardcoded per-script family-name specs with platform script fallbacks, deleting ~350 lines of brittle name-matching. This is the right shape for cross-platform font coverage; real-device confirmation from the reporter would be the ideal follow-up on #106.

Landing note: large diff in `src/system_fonts.rs`, but pairwise conflict-free against the rest of today's batch.

RichardCao:
## macOS validation

I have now validated the merged PR on a real Apple Silicon Mac:

- macOS 26.4.1 (Darwin 25.4.0), arm64, locale `zh_CN`
- Rust/Cargo 1.95.0
- exact merge commit: `132b728ba7375d809c9d825a7fd8eda71d57f2af`
- CoreText fallback probe: **PASS**
  - Simplified/Traditional Chinese: PingFang
  - Devanagari: SF Devanagari
  - Thai: Thonburi
  - Korean: Apple SD Gothic Neo
- `installed_fonts_cover_reported_scripts`: **PASS** (`1 passed; 0 failed`)
- application build: **PASS**
- visual check of regular and bold text: **PASS** for Simplified/Traditional Chinese, Japanese, Korean, Devanagari, and Thai; no tofu, missing combining marks, or clipping observed

This confirms that the Fontique/CoreText fallback path fixes the macOS glyph-coverage problem reported in #106 without hardcoded installed font family names. A screenshot and the full build log were retained locally and can be provided if useful.

PR #115, opened 2026-08-27
Balance fitted table columns by content demand

### LINKED ISSUE #112: Preserve relative content demand when fitting table columns
## Context

Follow-up to #71 and the table-wrapping implementation in #98.

The current `fit_column_widths` policy finds one common cap and applies `min(desired, cap)` to every column. This is stable and max-min fair, but when several columns exceed the cap it makes them equal even when their intrinsic widths differ substantially.

For example, desired widths `[200, 400, 300]` in 600 px become approximately `[200, 200, 200]`. The second column has twice the content demand of the first but receives no additional space.

## Desired invariants

- compact tables remain at natural width;
- every fitted column stays between its hard minimum and normalized natural width;
- fitted widths sum to the available width;
- greater natural demand never produces a narrower fitted column;
- resizing and small font-metric changes produce continuous width changes;
- one extreme outlier cannot starve all neighboring columns;
- horizontal scrolling remains the fallback when hard minimums cannot fit.

## Revised proposal after review

Use a continuous convex blend of two allocations:

1. **60% max-min fair allocation:** the existing capped/water-filling solution, which protects neighboring columns from an outlier.
2. **40% proportional-demand allocation above the hard minimum:** preserves the ordering and relative differences of natural widths.

Both components independently satisfy the hard bounds, total-width constraint, ordering, and continuity; their blend retains those properties. The 60% value therefore expresses an explicit fairness weight rather than acting as a discontinuous classification threshold.

The tradeoff is that a narrow column is not always preserved at its full natural width while fitting is active. That property conflicts in general with both preserving demand order and giving wider columns meaningfully different widths. Hard-minimum columns are still protected exactly.

Regression coverage should include boundary continuity, outliers, equal demands, below-minimum inputs, compact tables, total width, bounds, and demand ordering.

This issue is independent of Full Width semantics (#110), viewport clipping (#111), and cell row-height measurement.

RichardCao:
## Summary

- blend max-min fairness with proportional unmet content demand when table columns must fit
- reserve the widest Unicode word in each header as a per-column floor, so short labels such as `Required` are not split mid-word
- distribute only the width above those floors with exact sorted water-filling and `f64` intermediates
- apply the same floor and horizontal-overflow behavior to Markdown and HTML tables
- preserve hard minimums, natural-width bounds, continuity, and compact-table behavior; equal-floor allocations retain demand ordering

## Current base

- rebased cleanly onto current `main` at `759a585` (v0.2.0, including #131)
- #131's table-height reservation remains intact and uses the same computed initial widths

## Testing

- 13 focused allocation/header-floor tests cover continuity, unequal floors, compact tables, single columns, and extreme outliers
- production rendering regression reproduces the five-column `Name / Description / Allowed Types / Required / Games` case at 400 px and asserts `Required` remains one galley row; it fails when the header floors are replaced with the old 40 px floor
- renderer tests with all CI features: 76 library tests, 1 slice test, 19 wrapping tests, and 16 doc tests passed (1 doc test ignored)
- application: 60 tests passed (1 system-font test ignored)
- `cargo fmt --check`, root `cargo clippy -- -D warnings`, and `git diff --check` passed
- post-rebase high-level review found no blocker or semantic loss

Fixes #112

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1491,6 +1491,7 @@ dependencies = [
  "emojis",
  "pulldown-cmark 0.13.4",
  "resvg",
+ "unicode-segmentation",
 ]
 
 [[package]]
--- a/crates/egui_commonmark/Cargo.lock
+++ b/crates/egui_commonmark/Cargo.lock
@@ -1170,6 +1170,7 @@ dependencies = [
  "image",
  "pulldown-cmark 0.13.0",
  "resvg",
+ "unicode-segmentation",
 ]
 
 [[package]]
--- a/crates/egui_commonmark/Cargo.toml
+++ b/crates/egui_commonmark/Cargo.toml
@@ -28,6 +28,7 @@ egui_commonmark_macros_extended = { version = "0.28.0", path = "egui_commonmark_
 document-features = { version = "0.2" }
 
 pulldown-cmark = { version = "0.13", default-features = false }
+unicode-segmentation = "1.12"
 
 
 [patch.crates-io]
--- a/crates/egui_commonmark/egui_commonmark/Cargo.toml
+++ b/crates/egui_commonmark/egui_commonmark/Cargo.toml
@@ -29,6 +29,7 @@ emojis = { workspace = true }
 document-features = { workspace = true, optional = true }
 
 pulldown-cmark = { workspace = true }
+unicode-segmentation = { workspace = true }
 
 [features]
 default = ["load-images", "pulldown_cmark"]
--- a/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
@@ -13,6 +13,7 @@ use egui_commonmark_backend_extended::elements::{
 use egui_commonmark_backend_extended::misc::*;
 use egui_commonmark_backend_extended::pulldown::*;
 use pulldown_cmark::{CowStr, HeadingLevel};
+use unicode_segmentation::UnicodeSegmentation;
 
 /// Search-match highlight kind for a single rendered text segment.
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
@@ -412,6 +413,18 @@ fn natural_text_width(ui: &Ui, text: &str) -> f32 {
         + 16.0
 }
 
+/// Width needed to keep the widest Unicode word on one line.
+///
+/// Body content may wrap aggressively, but a short table header such as
+/// `Required` should not be forced into `Require` / `d`. Unicode word
+/// boundaries keep CJK headers breakable while treating shaped scripts and
+/// combining sequences as words rather than splitting scalar values.
+fn unbreakable_text_width(ui: &Ui, text: &str) -> f32 {
+    text.unicode_words()
+        .map(|word| natural_text_width(ui, word))
+        .fold(40.0, f32::max)
+}
+
 fn wrapped_text_height(ui: &Ui, text: &str, column_width: f32, line_height: f32) -> f32 {
     let font_id = egui::TextStyle::Body.resolve(ui.style());
     let galley = ui.painter().layout(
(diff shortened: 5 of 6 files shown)


aydiler:
Tested this locally, rebased onto current `main` (clean rebase, builds without warnings). The column balancing does what it says on documents where the description column was starved — tables come out visibly more compact.

But it regresses a case that `main` now renders correctly, so I'd like to hold the merge on this.

## Regression: a column can be squeezed below its header's word width

Same table (`Instances` → `Children`), same document (`asset_reference.md` from #121), same 1200x800 window, same scroll position:

**current `main`** — header row on one line:

```
| Name | Description                        | Allowed Types | Required | Games |
| core | For Deadlocked mission instance... | Instances     | No       | DL    |
```

**this branch** — `Required` is broken mid-word, and `Allowed Types` is forced to two lines:

```
| Name | Description               | Allowed | Require | Games |
|      |                           | Types   | d       |       |
| core | For Deadlocked mission... | Instances | No    | DL    |
```

`Require` / `d` is the same class of defect the reporter of #121 complained about in their second screenshot — a cell broken mid-word because its column is narrower than the content requires.

## Why this matters more than it might look

Worth flagging because it is easy to miss: **#113 already fixed the column clipping on this document.** On `7b8f53b` the Build-section table rendered `Collecti`/`on` and `Not yet documeni`/`ed.`; on current `main` the same table renders `Collection` and `Not yet documented.` intact. I believe that is 3fb66ea (stable table identity across slices) keeping column state from being recomputed per slice.

So the baseline this branch is balancing against is better than it was when the branch was written, and the demand-based redistribution now takes width away from columns that had just been fixed.

## Suggested constraint

A column's minimum should never go below the widest unbreakable token in its header. `Required` and `Allowed Types` are both header-driven minimums here; the balancing is free to redistribute everything above that floor.

## What is fine

- No scroll regression: deep-scroll probe on the same document shows zero blank frames, document bottom reached (at step 33 vs 35 on main — consistent with the more compact layout).
- Rebase onto `main` after #113 and #109 is conflict-free.

Happy to re-test once the header-width floor is in.

aydiler:
Heads-up: `main` has moved a fair way since this branch, and GitHub now reports the PR as conflicting.

Merged since you opened it: #113 (viewport slice correctness), #114 (Full Width table semantics), #109, #120, #126, plus #127–#129. The conflict is in `crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs` — the same file this PR reworks.

Two things worth knowing before you rebase, because they change what this PR is measured against:

- **#113 fixed the column clipping on its own.** The Build-section table in the #121 document rendered `Collecti`/`on` and `Not yet documeni`/`ed.` on `7b8f53b`; on `main` it renders `Collection` and `Not yet documented.` intact. I believe that is the stable-table-identity commit keeping column state from being recomputed per slice. So the starved-column baseline this branch balances against is better than it was.
- **#129 added an `options` parameter to `table_cell_height`** (to resolve image URIs during row measurement). If you touch that signature the rebase will want a hand there.

The blocker I noted earlier still stands separately: on this branch, `Required` breaks as `Require`/`d` in a table `main` renders on one line. A column's minimum should not fall below the widest unbreakable token in its header; above that floor the balancing is free to redistribute.

No urgency from my side — just did not want the conflict to be a surprise.

RichardCao:
Rebased this PR onto current main and addressed the header-word regression from the review.

Each column now reserves the widest Unicode word in its header before the remaining width is balanced. If those floors do not fit, the existing horizontal scroll path is used instead of splitting a short word mid-token. Both Markdown and HTML tables use the same rule.

I added a production rendering regression using the Name / Description / Allowed Types / Required / Games shape at 400 px. Required is asserted to remain one galley row; replacing the new floors with the previous 40 px floor makes the test fail. The allocation was also simplified to reserve floors first and run the existing sorted water-filling over remaining headroom.

High-level code review found no blocker, and the full local renderer/application suites passed. CI is running on 629e494.

aydiler:
The header-word fix works — verified on the document from #121 at the same position I reported it from. `Required` renders intact where it previously broke as `Require`/`d`, and `Allowed Types` wraps between its two words, which is exactly the rule you described. The eleven added tests all pass, the rebase is clean, and all three CI jobs are green.

**But I'm holding the merge on a different finding: this branch paints an empty document pane while scrolling.**

## What I measured

`scripts/scroll-regression.sh` (added in #126 — it walks a document in fine increments and fails on an empty document pane), default fixture, same session, same display:

| build | result |
|---|---|
| `main` @ 2cca28a | **PASS** — no blank frames, bottom reached at frame 103 |
| this branch | **FAIL** — `1 frame(s) painted an empty document pane: f011.png` |

Deterministic: two consecutive runs, same frame. The control run on `main` in the same session rules out timing or a broken probe — that check matters, since a clean run can otherwise just mean the blank moved out of the walk's path.

## Where it lands

At a boundary between two tables, which is the same signature as #125:

| frame | content |
|---|---|
| f010 | Section 2's table, row `entry_2_21` |
| **f011** | **empty document pane** — explorer and outline still paint |
| f012 | Section 3's table, row `entry_3_2` |

## Reproducing

```bash
scripts/scroll-regression.sh target/debug/md-viewer
# MDV_DISPLAY=98 if :99 is occupied
```

Requires Xvfb, xdotool, ImageMagick and python3 with Pillow. It is not wired into CI for exactly that reason — which is why all three jobs here are green while this is happening.

## A guess, not a diagnosis

The fix falls back to the horizontal scroll path when the header floors do not fit. That path is where the split-point problems behind #113 and #125 lived. First place I would look, but I have not confirmed it.

Happy to re-run the guard against a new push, or to dig into the cause myself if you would rather hand it over — the fix itself is right and I would like it in.

aydiler:
**Correction: the blank frame was not yours. I withdraw the hold.**

My control run was against `main` @ 2cca28a — and #131, the table height reservation, landed *after* that. So the comparison I based the hold on never contained the fix. I guessed at the horizontal-scroll path in that comment; I did not check the more obvious possibility, that the defect belonged to `main` rather than to this branch.

## Evidence

Rebased this branch onto current `main` (clean, no conflicts) and ran the identical guard — same session, same display, same fixture:

| build | f011 `content_px` | result |
|---|---|---|
| this branch on its original base (2cca28a, **no** #131) | **197** | FAIL — blank pane |
| this branch rebased onto `main` (**with** #131) | **1954** | PASS, bottom at frame 108 |

The neighbouring frames are essentially unchanged between the two runs — f010 1814 → 1811, f012 2176 → 2173, f013 1927 → 1924 — so the walk visits the same content and only f011 differs. That rules out the alternative reading, that the blank simply moved somewhere the walk no longer passes.

#131's root cause fits exactly: `egui_extras` reserves the height of skipped rows only once it reaches the first *visible* row, so a table lying entirely above the viewport reserved nothing and everything below it laid out too high.

The bottom moving from 103 to 108 frames is expected — this PR changes column widths, so the document's height changes with it.

## On the rebased branch

- root: 60 passed, 1 ignored
- renderer: 76 lib + 19 wrapping + 1 integration + 16 doctests passed
- `scripts/scroll-regression.sh`: PASS

`maintainerCanModify` is on, so I could push the rebase — but I'd rather not force-push your branch out from under you. Rebasing on current `main` is all it needs; say the word and I'll push it, or push it yourself and I'll merge.

Sorry for the detour. The fix itself was right the whole time.

RichardCao:
Rebased onto current main at 759a585 (including #131 and the v0.2.0 update) and force-pushed as 13be977. The rebase was clean. Full renderer/application tests and a post-rebase high-level review passed; #131 table-height reservation remains intact. CI is running on the updated head.

aydiler:
Verified `13be977` independently — CI does not run the scroll guard, and that guard is what produced my incorrect hold in the first place, so I did not carry over the result from my own rebase.

| | |
|---|---|
| `scripts/scroll-regression.sh` | **PASS**, bottom at frame 108 |
| f011 — the frame that was blank | `content_px` **1955** (197 on the old base) |
| `table_cell_height` image + math arms | intact |
| CI check / renderer / version-sync | pass |

Your branch sits on `759a585` rather than current `main` — the two commits since are a LESSONS entry and the crates.io token pre-flight, no code, so there is nothing to rebase for.

Merging. Thanks for the quick turnaround, and sorry again for the false alarm.

Issue #71, opened 2026-08-22
Long lines in tables are not wrapped


h8ghbzc9bbr8c7:
Found in 0.1.16: When tables columns contain long lines which do not fit in the column, they are not wrapped (split in more than one displayed line). Instead a scroll bar is displayed.

I Would expect that (at least as default) lines which do not fit in the tables column are wrapped.

RichardCao:
I looked into this and the behavior comes from the current table layout rather than a platform-specific issue. Markdown and HTML tables use content-sized `Column::auto()` columns inside a horizontal `ScrollArea`, so a long cell expands its column and produces horizontal scrolling instead of wrapping.

A proper fix should do more than enable a wrapping layout: table row heights are currently calculated before rendering, so simply wrapping labels can clip the additional lines behind the next row or border. The likely implementation path is to:

- preserve natural widths for compact tables;
- distribute the available table width across columns when their desired total is too wide;
- wrap ordinary cell text at the resulting widths;
- calculate heterogeneous row heights from the actual column widths;
- retain horizontal scrolling when column minimum widths still exceed the pane, or when the user explicitly expands columns;
- apply the same policy to Markdown and HTML tables.

It would be better to implement this after #91 and #76 are merged:

- #91 adds the vendored renderer workspace to CI, which is important because this fix needs renderer-level regression coverage.
- #76 changes the same table row-height calculation for inline math. Building the wrapping calculation on top of that result avoids conflicting implementations and preserves formula-height handling.

Once those PRs are confirmed merged, I can prepare a focused PR for this issue against the latest `main`, with regression tests for wrapped prose, dynamic row height, compact-table sizing, resizable columns, and the horizontal-scroll fallback.

PR #116, opened 2026-08-27
Fix wrapping and row heights for mixed table content

RichardCao:
## Summary

- measure inline code with the Body-sized monospace font used by rendering
- measure text, links, inline HTML, footnote references, and inline code in event order with production-like wrapping
- isolate each cell wrapping cursor from TableBuilder preallocated row height so later inline widgets cannot grow past the row clip
- derive table row height from configured typography instead of a separate hard-coded multiplier
- reserve cumulative height for inline math, images, and mixed content using the current renderer options
- discard structural `End(TableCell)` events instead of painting them as hidden spacing in the next cell

## Regression coverage

Production geometry tests cover:

- multiple links wrapping within one narrow cell, including horizontal and vertical clip bounds and following content
- Chinese prose mixed with `raw_formula`, `hash`, and `formula_status=unclear` inline code
- text before and after inline math with the `math` feature enabled
- a representative 13-column selected-family table
- the application line-height configuration

Additional unit tests cover Body/Monospace width differences, multiple chunked code events, remaining-width wrapping, configured typography resolution, and parsed cell boundaries.

## Rebase integration

Rebased onto current `main` after #115 merged. The conflict resolution preserves:

- #114 table bounds and resize reflow
- #115 header floors and balanced column widths
- #129 image measurement and renderer options
- #131 exact reserved table height

The #129 options adaptation is folded into the first functional commit, leaving three focused commits.

## Testing

- application: 60 passed, 1 environment-dependent font test ignored
- renderer library with math: 74 passed
- renderer slice/wrapping integration with math: 1 + 23 passed
- renderer doctests with math and macros: 16 passed, 1 ignored
- backend with all features: 28 passed, 1 doctest passed
- macros: 4 unit tests and 7 trybuild cases passed; 1 doctest passed, 2 ignored
- `git diff --check`
- independent high-level review of the final rebased head: no blocker

Follow-up to #71 and #98.

### DIFF
--- a/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
@@ -311,11 +311,10 @@ fn inline_code_wrap_segments(text: &str) -> Vec<String> {
     segments
 }
 
-/// Measure a table cell's visible labels in their rendering order. Each label
-/// is laid out against the width left by earlier labels in the current row;
-/// this matters when a wider monospace label follows proportional text. Long
-/// code chunks call `ui.end_row()` in production, so they also finish a
-/// measurement group here.
+/// Measure a table cell's visible labels in their rendering order. The cell
+/// uses a wrapping horizontal layout: every label continues the current row,
+/// then uses the full column width for subsequent rows. Long code chunks call
+/// `ui.end_row()` in production, so they also finish a measurement row here.
 fn cell_visual_lines(
     cell: &[(pulldown_cmark::Event, Range<usize>)],
     ui: &Ui,
@@ -329,31 +328,33 @@ fn cell_visual_lines(
     let wrap_width = (column_width - 8.0).max(1.0);
     let item_spacing = ui.spacing().item_spacing.x;
     let mut remaining_width = wrap_width;
-    let mut group_lines = 0usize;
-    let mut visual_lines = 0usize;
-    let measure_label = |text: &str, font_id: &egui::FontId, available: f32| {
-        ui.painter().layout(
+    let mut completed_lines = 0usize;
+    let measure_label = |text: &str, font_id: &egui::FontId, used_width: f32| {
+        let mut job = egui::text::LayoutJob::simple(
             text.to_owned(),
             font_id.clone(),
             egui::Color32::WHITE,
-            available.max(1.0),
-        )
+            wrap_width,
+        );
+        if let Some(first_section) = job.sections.first_mut() {
+            first_section.leading_space = used_width;
+        }
+        ui.fonts_mut(|fonts| fonts.layout_job(job))
     };
 
     let add_label = |text: &str,
                      font_id: &egui::FontId,
                      remaining_width: &mut f32,
-                     group_lines: &mut usize| {
-        let galley = measure_label(text, font_id, *remaining_width);
-        let lines = galley.rows.len().max(1);
-        *group_lines = (*group_lines).max(lines);
-        // A wrapped Label consumes the available horizontal slot. A one-line
-        // Label advances only by its actual width plus inter-widget spacing.
-        if lines > 1 {
-            *remaining_width = 1.0;
-        } else {
-            *remaining_width = (*remaining_width - galley.size().x - item_spacing).max(1.0);
-        }
+                     completed_lines: &mut usize| {
+        let used_width = wrap_width - *remaining_width;
+        let galley = measure_label(text, font_id, used_width);
+        *completed_lines += galley.rows.len().saturating_sub(1);
+        let last_row_width = galley
+            .rows
+            .last()
+            .map_or(0.0, |row| row.rect().width())
+            .min(wrap_width);
+        *remaining_width = (wrap_width - last_row_width - item_spacing).max(0.0);
     };
 
     for (event, _) in cell {
@@ -362,7 +363,7 @@ fn cell_visual_lines(
             | pulldown_cmark::Event::InlineHtml(text)
             | pulldown_cmark::Event::Html(text)
             | pulldown_cmark::Event::FootnoteReference(text) => {
-                add_label(text, &body_font, &mut remaining_width, &mut group_lines);
+                add_label(text, &body_font, &mut remaining_width, &mut completed_lines);
             }
             pulldown_cmark::Event::Code(code) => {
                 let segments = inline_code_wrap_segments(code);
@@ -372,31 +373,30 @@ fn cell_visual_lines(
                         &segment,
                         &code_font,
                         &mut remaining_width,
-                        &mut group_lines,
+                        &mut completed_lines,
                     );
                     if force_rows {
-                        visual_lines += group_lines;
-                        group_lines = 0;
+                        completed_lines += 1;
                         remaining_width = wrap_width;
                     }
                 }
             }
             pulldown_cmark::Event::SoftBreak | pulldown_cmark::Event::HardBreak => {
-                add_label(" ", &body_font, &mut remaining_width, &mut group_lines);
+                add_label(" ", &body_font, &mut remaining_width, &mut completed_lines);
             }
             pulldown_cmark::Event::TaskListMarker(checked) => {
                 add_label(
                     if *checked { "[x] " } else { "[ ] " },
                     &body_font,
                     &mut remaining_width,
-                    &mut group_lines,
+                    &mut completed_lines,
                 );
             }
             _ => {}
         }
     }
 
-    (visual_lines + group_lines).max(1)
+    (completed_lines + usize::from(remaining_width < wrap_width)).max(1)
 }
 
 fn table_cell_height(
@@ -409,8 +409,22 @@ fn table_cell_height(
 ) -> f32 {
     let text = markdown_cell_text(cell);
     let mut height = wrapped_text_height(ui, &text, column_width, line_height).max(
-        line_height * 1.5 * cell_visual_lines(cell, ui, column_width) as f32,
+        line_height * cell_visual_lines(cell, ui, column_width) as f32
+            + ui.spacing().item_spacing.y,
     );
+    let has_visible_text = cell.iter().any(|(event, _)| {
+        matches!(
+            event,
+            pulldown_cmark::Event::Text(_)
+                | pulldown_cmark::Event::Code(_)
+                | pulldown_cmark::Event::InlineHtml(_)
+                | pulldown_cmark::Event::Html(_)
+                | pulldown_cmark::Event::FootnoteReference(_)
+                | pulldown_cmark::Event::TaskListMarker(_)
+        )
+    });
+    let mut inline_math_height = 0.0;
+    let mut inline_math_count = 0usize;
     for (event, _) in cell {
         // Images contribute their painted height. The URI has to be resolved
         // exactly the way the painting path does it (`Image::new` applies the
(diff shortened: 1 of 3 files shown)


RichardCao:
Added a follow-up fix for another row-clipping case found in a real 13-column report table.

`parse_row` was retaining each `End(TableCell)` boundary as the first event of the next cell. The renderer then painted that structural event as a two-space label, so production layout had an extra ~6 px first-line indent that the row-height measurement did not see. Near a wrap boundary this added one painted line and clipped the final line (for example, the tail of a confidence interval).

Commit `5b4f334` now discards the cell boundary after closing the preceding parsed cell. TableBuilder/Grid already provides column separation, so no visible spacing or interaction is lost. This also slightly reduces unnecessary wrapping rather than padding every affected row.

Regression coverage:
- backend test verifies parsed cells contain no neighboring `End(TableCell)` boundary;
- a production-rendering 13-column fixture reproduces the selected-family table geometry and asserts no text crosses its vertical clip.

High-level review found no blocker. Verified backend tests (15 + 1 doc), renderer wrapping tests (23), macro tests/trybuild, and the application suite locally.

aydiler:
Looked at the red `renderer` job. Three things, in order of what I think matters most.

## 1. The failing test is one this PR adds, and it fails locally too

`markdown_table_wraps_multiple_links_inside_their_cell` does not exist on `main`, so this is not an external regression — the implementation here does not yet satisfy the case the branch itself introduces.

Reproduced locally with byte-identical numbers to CI:

```
marker "REPORT_IC" is clipped:
  rect      = [[79.6 40.5] - [294.5 60.5]]   width 214.9
  clip_rect = [[79.6 39.0] - [291.0  84.0]]  width 211.4
```

3.5 px over the right edge, on the third of four links in one cell at 280 px. Deterministic, not a CI flake. The other 19 tests in the file pass.

One observation, offered as a lead rather than a diagnosis: `cell_visual_lines` computes how many *lines* a cell needs, but this failure is *horizontal*. Links are accumulated into `self.link` and emitted as a unit at `TagEnd::Link`, so a link whose label does not fit the width left on the current row has no obvious place to break. The assertion that fails is `assert_text_fully_visible`, which runs before your `link_tops.len() > 1` check, so the run does not actually tell us yet whether the link *wrapping* works — only that one label lands outside the clip.

## 2. `main` has moved a long way under this branch

The branch is 22 commits behind, and two of those merges are directly under this PR's feet:

- **#113** — stable table identity across viewport slices. This turned out to fix the column clipping in #121 on its own: the Build-section table rendered `Collecti`/`on` and `Not yet documeni`/`ed.` on `7b8f53b` and renders them intact on `main`.
- **#114** — tables now follow the Full Width setting, so with Full Width off a table is capped at the reading width (measured 794 px → 599 px on a 1600 px window) instead of taking the centre pane.

Both change the widths this PR measures against. Worth re-measuring the motivating cases on current `main` before investing more here — some of them may already behave differently.

## 3. The rebase now conflicts

`git rebase origin/main` stops on `crates/egui_commonmark/egui_commonmark/tests/wrapping.rs`. #114 added `table_cell_height_after_widths` and `markdown_table_reflows_after_panel_width_changes` in the same region where this branch adds its helpers. The conflict is additive on both sides, so it should resolve cleanly by hand — flagging it so it is not a surprise.

## Sequencing with #115

#115 and this PR both edit `parsers/pulldown.rs`, so whichever lands first forces a rebase on the other. #115 has its own blocker (I left a note there: it squeezes a column below its header's word width, breaking `Required` as `Require`/`d` on a table `main` renders correctly). No rush on ordering from my side — just be aware they will collide.

aydiler:
Heads-up on top of the earlier review: `main` has moved and GitHub now reports this PR as conflicting.

Conflicting files: `crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs` and `crates/egui_commonmark/egui_commonmark/tests/wrapping.rs`.

Merged since you opened it: #113, #114, #109, #120, #126, #127–#129. The `wrapping.rs` conflict is the one I flagged before — #114 added `table_cell_height_after_widths` and `markdown_table_reflows_after_panel_width_changes` in the same region where this branch adds its helpers. Additive on both sides, so it should resolve by hand cleanly.

One more since then: **#129 added an `options` parameter to `table_cell_height`** and an image arm to the row-height estimator, which is the same function this PR reworks. Worth reading that diff before rebasing rather than after.

The original blocker is unchanged: `markdown_table_wraps_multiple_links_inside_their_cell` — a test this branch introduces — still fails locally with the same numbers as CI (`REPORT_IC` at x=294.5 against a clip ending at 291.0).

aydiler:
I rebased this onto current `main` locally to see how bad the conflict actually is. It is small, but a clean rebase is **not** enough here — the compiler finds two collisions git cannot see.

## The textual conflict is one hunk, in a test file

`tests/wrapping.rs`: both sides append independent tests at the same spot. `main` gained `table_cell_height_after_widths` plus two reflow tests from #114; this branch adds three wrapping tests. Keep both.

One trap in it: the last function on the `main` side has its closing brace *after* the `>>>>>>>` marker, in the shared tail. Concatenating the two sides leaves that function open — `error: this file contains an unclosed delimiter`. I hit exactly that on the first attempt.

## Two signature collisions the rebase does not surface

Both are "the same function grew a parameter on each side, at different lines", so there is no textual conflict and the build fails afterwards:

- `parsers/pulldown.rs:2806` — this branch calls `table_cell_height` with 5 arguments; on `main` it takes 6 since #129 added `options: &CommonMarkOptions` for image heights. The call site is in your `table_cells_measure_inline_code_with_the_monospace_font` test, so `options` is not in scope — it needs a local `CommonMarkOptions::default()`, matching how the neighbouring height tests do it.
- `tests/wrapping.rs:218` — mirror image: you gave `collect_painted_text` a `clip_rect` parameter, and `main`'s reflow helper from #114 still calls the 2-argument form. Pass `clipped.clip_rect`.

## What I checked, since #114 and #131 both touched row heights

The thing I most wanted to rule out was silent loss. `table_cell_height`'s image arm (#129) and math arm both survive your rewrite of `cell_visual_lines` intact — compared line by line against `main`.

And the PR is clearly **not** redundant after #114/#131: it replaces the old heuristic (count `inline_code_wrap_segments` chunks, take the max) with real galley measurement across mixed cell content — text, inline HTML, footnote references, code — accumulating across the wrapped row. Orthogonal to #114 (table width bound) and #131 (the table reserving its own height).

## Results on the rebased branch

| | main | this branch rebased |
|---|---|---|
| lib tests | 65 | **69** (4 added, none lost) |
| wrapping tests | 19 | **22** (3 added) |
| doctests / integration | 16 / 1 | 16 / 1 |
| `scripts/scroll-regression.sh` | PASS, bottom at frame 103 | **PASS, bottom at frame 82** |

The document getting *shorter* — 82 frames instead of 103 — is the fix doing its job: rows are no longer over-allocated.

`cargo fmt` clean.

I have this rebase sitting locally and `maintainerCanModify` is on, but the two signature fixes change your code, so they should go in under your review rather than mine. Say the word and I'll push it, or take the four line numbers above and push it yourself — either way I'd like this in.

RichardCao:
Thanks for the detailed rebase notes. #115 has now been rebased onto current main, high-reviewed, and is green; I am preparing this branch locally on a synthetic base that assumes #115 is merged. I will preserve both sides of the wrapping tests, adapt the #129 options parameter and #114 clip_rect call, and re-check the multi-link regression plus the #129 image and math paths. I will not force-push this PR until #115 actually lands, so its published base stays truthful and we only rebase once.

aydiler:
#115 is merged as `f122909e`. Its published base is now real, so this one is clear to rebase — your plan to wait for it was the right call.

Before merging I verified `13be977` with `scripts/scroll-regression.sh`, which CI does not run: PASS, bottom at frame 108, and f011 — the frame that was blank on the old base — back to `content_px` 1955.

When you push this branch, I will run the same guard on it rather than reusing my own trial rebase, and check the #129 image and math arms the way you listed.

RichardCao:
Rebased onto current `main` now that #115 is merged and force-pushed as `8ecdde3`.

I preserved the additive #114 wrapping/reflow tests and fixed both non-textual signature collisions from the review: the #129 `options` argument and the #114 `clip_rect` argument. I also rechecked that #115 header floors, #129 image/math measurement, and #131 reserved table height remain intact.

The previously failing multi-link regression now passes and verifies all four link markers stay inside their horizontal and vertical clip rects, retain link styling, wrap across multiple rows, and do not overlap following content.

Local verification: application 60 passed/1 ignored; renderer 74 library + 1 slice + 23 wrapping; renderer doctests 16/1 ignored; backend 28 + 1 doc; macros and trybuild all passed. The exact final head received an independent high-level review with no blocker.

Please rerun the scroll guard as planned; this environment does not currently have the Xvfb/xdotool/ImageMagick harness dependencies installed.

aydiler:
Ran the verification I promised, against current `main` — which now includes #115, #136, #137 and #138, so the numbers differ from my earlier trial rebase.

Rebased your branch onto `main` and applied the two signature fixes from my previous comment:

| | |
|---|---|
| lib tests | **80** |
| wrapping tests | **23** |
| integration / doctests | 1 / 16 (1 ignored) |
| `scripts/scroll-regression.sh` | **PASS**, bottom at frame 85 |

**The test your CI is currently red on passes.** `markdown_table_wraps_multiple_links_inside_their_cell` panicked at `wrapping.rs:103` in [run 33783140506](https://github.com/aydiler/md-viewer/actions/runs/33783140506) yesterday; on current `main` it runs green — checked by name, not inferred from a green suite:

```
test markdown_table_wraps_multiple_links_inside_their_cell ... ok
```

So the red CI you are looking at is stale, not a defect to chase.

One warning from doing it: **the conflict is three hunks now, not one.** #115 added its own tests to `wrapping.rs`, and two of the new hunks are inside `PaintedText` — `main` contributes `rows: usize`, your branch `clip_rect: Rect`, and both the struct and its initializer need *both* fields. I applied my earlier "keep both sides plus a closing brace" recipe blindly, closed the struct early, and got `error: unexpected closing delimiter`. Brace balance was still 0, so that check does not catch it — it counts braces, not placement.

Nothing here needs anything from me. Push when you are ready and I will re-run the guard on your commit rather than reusing this trial.

aydiler:
Ran the guard on `8ecdde3` as asked — your exact head, not my trial rebase, which did not contain the `parse_row` fix.

| | |
|---|---|
| `scripts/scroll-regression.sh` | **PASS**, bottom at frame 85 of 150 |
| blank frames | 0 |
| stuck scrolling | none before the bottom |

Also checked, by inspection rather than on your word, that nothing was lost in the rebase — this is the place where a silent loss would be most expensive and the one thing you could not verify without the harness:

| | |
|---|---|
| #129 `options` parameter and image arm in `table_cell_height` | present |
| #129 math arm | present |
| #115 header floors | present |
| #131 reserved table height | present |

The bottom moving 108 → 85 is the expected direction: with cell content measured properly, rows stop being over-allocated and the document gets shorter.

One note on `parse_row` retaining each `End(TableCell)` as the next cell's first event, painted as a two-space label: that is a much better explanation of yesterday's red CI than "the rebase fixed it", which is what I would have concluded from a green suite. Worth having in `docs/LESSONS.md` — happy to write it up, or leave it to you if you would rather phrase it yourself.

Your branch sits on `main` from just before #145 merged, which is a documentation-only commit. Nothing to rebase for. Merging.

PR #119, opened 2026-09-01
Add font selection UI for customizable document font

Pat9496:
## Summary

Allows users to browse and apply installed system fonts via View menu. Font selection is searchable, cached for performance, and persists across restarts. Users can now choose their preferred reading font instead of relying solely on the auto-detected system sans-serif.

## Test Plan

- [ ] Open View → Font to verify picker renders
- [ ] Search for a font name and verify filter works
- [ ] Select a font and verify document updates immediately
- [ ] Restart the app and verify font selection persists
- [ ] Verify monospace/code font is unchanged
- [ ] Verify performance with many fonts (picker remains responsive)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ A fast, lightweight markdown viewer for Linux built with Rust and egui. Designed
 - **Keyboard Scrolling** - Scroll documents with ↑/↓ by line or Page Up/Page Down by page when the find bar is closed
 - **Live Reload** - Auto-refresh on file changes (enabled by default)
 - **Custom Colors** - Customize highlight and link text colors (View → Colors…)
+- **Font Selection** - Browse and apply installed system fonts via View menu (searchable, persistent)
 
 *Full Width, Formula Size and the colour picker all live in the View menu*
 
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ A fast, lightweight markdown viewer for Linux built with Rust and egui. Designed
 - **Live Reload** - Auto-refresh on file changes (enabled by default)
 - **Custom Colors** - Customize highlight and link text colors (View → Colors…)
 - **Font Selection** - Browse and apply installed system fonts via View menu (searchable, persistent)
+- **Smooth Text Rendering** - Optional toggle (View → Smooth Text Rendering) that disables glyph pixel-snapping for less jagged-looking text at fractional zoom levels
 
 *Full Width, Formula Size and the colour picker all live in the View menu*
 
--- a/docs/devlog/056-font-selection.md
+++ b/docs/devlog/056-font-selection.md
@@ -1,0 +1,142 @@
+# Feature: Document Font Selection
+
+**Status:** ✅ Complete
+**Branch:** `feature/font-selection`
+**Date:** 2026-08-31
+**Lines Changed:** +226 / -29 across `src/main.rs` (+134/-24) and `src/system_fonts.rs` (+121/-5)
+
+## Summary
+
+Lets the user pick which installed system font renders the markdown body/heading
+text (the `FontFamily::Proportional` chain), instead of always using the
+auto-detected system sans-serif. Selection persists across sessions. The
+monospace/code font is untouched.
+
+## Features
+
+- [x] `system_fonts::setup_fonts` accepts an optional preferred family name and
+      falls back to today's auto-detected sans-serif chain when unset/not found
+- [x] `setup_fonts` returns the sorted list of installed family names so the UI
+      doesn't need a second `fontique::Collection` scan
+- [x] `View → Font…` opens a searchable picker window (`egui::Window`)
+- [x] Selection persisted in `PersistedState.selected_font_family`
+- [x] Font reload only runs when the selection actually changes (mirrors the
+      existing `last_applied_dark_mode` pattern), not per-frame
+
+## Key Discoveries
+
+### fontique name lookup is case-insensitive and safe to call every time
+
+`Collection::family_id(&mut self, name: &str) -> Option<FamilyId>` matches
+against a lowercased key (`fontique-0.11.1/src/family_name.rs`), so any name
+returned from `family_names()` round-trips through `family_id()` without
+needing exact-case bookkeeping.
+
+### Bold companion resolution needed no changes
+
+`install_strong_font_family` already derives the bold face from whichever
+family was installed as `primary` (looked up by `regular.selected.family_id`),
+regardless of whether that family came from the auto-detect path or a
+user preference. Wiring the preferred family into the *same* `"SystemSans"`
+primary slot in `install_regular_fonts` meant the existing strong-font-family
+and script/generic fallback logic needed zero changes.
+
+## Architecture
+
+### New/Modified Structs
+
+```rust
+// PersistedState
+selected_font_family: Option<String>, // None = system default (auto-detected)
+
+// MarkdownApp
+selected_font_family: Option<String>,
+available_font_families: Vec<String>,   // populated once at startup from setup_fonts()'s return value
+last_applied_font_family: Option<String>,
+show_font_dialog: bool,
+font_filter: String,                    // transient UI-only search text, not persisted
+```
+
+### New Functions
+
+| Function | Purpose |
+|----------|---------|
+| `system_fonts::setup_fonts(ctx, preferred_family: Option<&str>) -> Vec<String>` | Same as before, plus optional named-family preference; now returns the sorted family name list |
+| `MarkdownApp::render_font_settings(&mut self, ctx)` | Renders the `View → Font…` picker window |
+
+## Testing Notes
+
+- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test`
+  (56 passed) all clean.
+- Added two new `system_fonts` tests behind `#[ignore = "requires installed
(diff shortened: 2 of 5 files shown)


aydiler:
I've approved the workflow runs, so CI has results now. Two of the three jobs are green — `renderer` and `version-sync` pass. The `check` job fails on one thing only:

```
cargo fmt --check
Diff in src/main.rs:4804:
-                        Some(if self.smooth_text_rendering { "on" } else { "off" }),
+                        Some(if self.smooth_text_rendering {
+                            "on"
+                        } else {
+                            "off"
+                        }),
```

Running `cargo fmt` and pushing should turn it green — nothing else is failing.

One heads-up on filenames while you're in there: this branch adds `docs/devlog/056-font-selection.md` and `057-smooth-text-rendering.md`. Both numbers are fine on your branch, but `main` has moved since you opened this, so a renumber may be needed at merge time. Not urgent, and not something to fix before CI is green.

Thanks for the contribution — I'll do a proper review of the font selection itself once CI is clean.

aydiler:
Two things, both from `main` moving rather than anything you did.

**1. This PR now conflicts** — in `src/main.rs`. Your other PR, #120 (highlight and link colours), merged in the meantime and touches the same area of the app struct and the View menu. Additive on both sides, so a rebase should resolve it by hand without trouble.

Also merged since you opened this: #113, #114, #109, #126, and #127–#129. Of those, **#130 is still open and adds a `View → Formula Size` submenu**, so if it lands before this one there will be a second small conflict in the same menu block. Nothing to do about it now, just so it is not a surprise.

**2. The `check` job is still red on the one formatting nit** from before:

```
cargo fmt --check
Diff in src/main.rs:4804:
-                        Some(if self.smooth_text_rendering { "on" } else { "off" }),
+                        Some(if self.smooth_text_rendering {
+                            "on"
+                        } else {
+                            "off"
+                        }),
```

`cargo fmt` and push fixes it; `renderer` and `version-sync` are already green.

While you are rebasing: this branch adds `docs/devlog/056-font-selection.md` and `057-smooth-text-rendering.md`, and both numbers are now taken on `main` (058 through 061 exist too). Renaming to 062/063 avoids a duplicate.

Thanks for the patience — #120 was a clean merge and I am looking forward to reviewing this one properly once CI is green.

aydiler:
I rebased this onto current `main` locally. It resolves, and the result is green — but two of the eight conflicts needed structural surgery rather than a merge, so please read those before trusting my resolution.

## Six are pure parallel additions

#120 (colors) and this PR (fonts) are the same shape of feature touching the same places: a field in `PersistedState`, a field in the app struct, initialization in the constructor, serialization back out, and the dialog call in `update()`. Keep both sides, `main` first.

## Two are not, and "keep both" produces broken code

**The two dialog functions.** git aligned `render_color_settings` and `render_font_settings` because they share an identical prologue — `if !flag { return; }`, `let mut open = true;`, and the comment above it. That shared text belongs to *each* function, so it has to be duplicated, and the shared tail that closes the function does too. I emitted the two complete functions separately rather than interleaving them.

**The View menu button.** Same trap: both sides open an `if btn.clicked() {` whose `ui.close(); }` sits in the shared tail below the marker. Each button needs its own copy, otherwise the first `if` never closes.

I mention the mechanism because I made exactly this mistake an hour earlier on #116 and got an `unclosed delimiter` out of it. Worth a second pair of eyes on my version.

## Results on the rebased branch

| | main | this branch rebased |
|---|---|---|
| app tests | 61 | **63** (2 added, none lost) |
| `cargo fmt --check` | clean | clean |
| `cargo clippy` | clean | clean |
| `scripts/scroll-regression.sh` | PASS, bottom at frame 103 | **PASS, bottom at frame 103** |

The two added tests are `system_fonts::known_preferred_family_becomes_primary` and `unknown_preferred_family_falls_back_to_auto_detect`. Both are skipped here because they need installed fonts, same as the existing CJK test — the ignored count goes 1 → 3 for that reason, not because anything was dropped.

The README conflict was one line: your **Font Selection** bullet against the **Custom Colors** bullet #120 added. Both kept.

`maintainerCanModify` is on, so I can push this rebase — but since two of the resolutions restructured your code rather than just merging lines, I would rather you look at them first. Say the word and I'll push, or take the description above and redo it yourself if you prefer your own hand on it.

aydiler:
Status check, three days on — and one correction to my own last comment, which has gone stale.

## Where this stands

- `check` is still red, on **one** thing: `cargo fmt --check`, the `smooth_text_rendering` block at `src/main.rs:4804`. `renderer` and `version-sync` pass.
- The branch is now **73 commits behind `main`**.

## My rebase resolution from 2026-09-04 is out of date — do not rely on it as written

I rebased again against current `main` just now to check rather than assume. Two findings:

**The 8 `src/main.rs` conflicts are unchanged**, and my earlier description of them still applies. They sit at:

| line | what |
|---|---|
| 378 | `math_scale: Option<f32>` in `PersistedState` |
| 1798 | `math_scale: f32` on the app struct |
| 2029 | `math_scale` in the constructor |
| 3946 | `fn render_color_settings` — one of the two needing real merging |
| 3958 | the deferred-action comment |
| 4390 | `math_scale: Some(self.math_scale)` in serialization |
| 4955 | the View-menu button block — the other one needing real merging |
| 5265 | the colour-picker window |

Six are parallel additions where "keep both, `main` first" is right. The two structural ones are the dialog function and the menu-button block, exactly as described before: each side's `if` block owns its own closing brace, and the shared prologue belongs to *each* function rather than being shared — "keep both sides" there produces an unclosed delimiter. Check `count('{') - count('}')` on the resolved file before compiling.

**There is now a ninth conflict, in `README.md`, and it is my fault, not yours.** I restructured the README in #162 and retook its screenshots in #168, which moved the section your bullet lands in. It is trivial to resolve — keep `main`'s block in full and add your line to the feature list above the screenshot:

```markdown
- **Custom Colors** - Customize highlight and link text colors (View → Colors…)
- **Font Selection** - Browse and apply installed system fonts via View menu (searchable, persistent)

*Full Width, Formula Size and the colour picker all live in the View menu*

![The View menu](screenshots/view-menu.png)
```

## What I need from you

`cargo fmt`, rebase onto current `main`, push. That is the whole remaining list.

I have deliberately not pushed a rebase to your branch — rewriting a contributor's history without asking is not something I will do unprompted. If you would rather I did, say so and I will; equally, if you would rather I opened a superseding PR carrying your commit with authorship intact, that also works. Your call.

Worth saying plainly: nothing here is a problem with the change itself. It is 555 lines that have been waiting on mechanical work while `main` moved underneath it, and the longer it waits the more it accumulates. Whichever of the three routes is least effort for you is the right one.

aydiler:
Checking in — no pressure, just so this does not sit unclear.

The branch is now **16 commits behind `main`** (v0.2.0 shipped, the vendored renderer moved to 0.28.1, and several link and table changes landed since), so the conflict resolution I described earlier has gone stale a second time. Nothing about the feature itself is in question.

What still applies from that earlier comment, in case it saves you time:

- eight conflicts against #120's colour feature; six are pure parallel additions,
- the two that need care are the dialog functions, which share a prologue (`if !flag { return; } let mut open = true;`) that belongs to **each** function and must be duplicated rather than shared,
- same trap in the View-menu button blocks: when both sides of a hunk end mid-block, the closing brace lives in the shared tail below the marker, so "keep both sides" leaves the first block unclosed. `count('{') - count('}')` on the resolved file catches it before the compiler does.

I have a working rebase locally that reached a clean build, tests, clippy and fmt, and the scroll guard green. I have deliberately not pushed it to your branch: restructuring your code under your name should be your call, not something that happens to you.

So — three ways forward, whichever suits you:

1. you rebase and push, and I re-run the guard on **your** commit (I would not reuse my local rebase for that; it has to be what you actually pushed);
2. you tell me to push my rebase to your branch and you review the result;
3. you would rather not continue, and I close this with the work credited — no hard feelings, and the colour dialog you contributed in #120 is shipped and in use.

If none of these gets a reply, I will leave the PR open rather than close it, but I will stop assuming it is in flight.

aydiler:
Rebased onto current `main` so this is reviewable again — the branch had gone stale against eight days of renderer work (#191, #194, #196, #198).

**What the rebase touched.** All eight conflicts were unions, not choices: `PersistedState` and `App` gained your font fields alongside main's `math_scale`/color-override fields; the View menu now carries both the **Colors…** button and your **Font: …** button (each closing the menu on click); `update()` renders both dialogs; the save/restore paths persist both feature sets; the README View section lists both. `render_color_settings` and `render_font_settings` now live side by side unchanged in substance. `system_fonts.rs` and both devlogs applied clean.

**Two small things I did while rebasing:**
- `cargo fmt` on `src/main.rs` — the Smooth-Text menu block had one hunk rustfmt would have failed CI on (applied as a fixup into your second commit, so the history stays two commits).
- The two `system_fonts` tests join the existing CJK one as `#[ignore]`d in CI (no fonts installed on the runner) — same pattern as before, they run locally where fonts exist.

**Verified locally:** builds clean, application suite 67 passed (3 ignored = the font-dependent tests), `cargo fmt --check` clean, `clippy -- -D warnings` clean, workspace sync check passes.

One review note: `available_font_families_lower` is a nice touch — the precomputed-lowercase cache means the picker's filter never re-lowercases per frame. CI will run on the updated branch; the rebase is yours to reject if the resolution order in the View menu isn't what you intended.

PR #120, opened 2026-09-01
Add selection highlight and link color customization

Pat9496:
## Summary

Adds a **View → Highlight Color…** color picker for customizing the selection highlight color (used in tabs, list entries, text selection, and other UI elements). 

- Color persists across restarts via session persistence
- Stored as premultiplied RGBA bytes (Color32's canonical format)
- "Reset to Default" clears the override and uses the current theme's stock color
- Picker is a separate window (learned that popups nested directly in menu_button get swallowed)

## Test Plan

- [ ] Open View menu → Highlight Color… opens a small window
- [ ] Click the color swatch to open the full RGBA color picker
- [ ] Drag the color picker: tab backgrounds and list selections change live
- [ ] Click "Reset to Default" to revert to the theme's stock blue
- [ ] Close and restart: selected color persists
- [ ] Change theme (Ctrl+D): custom color is still applied (overrides both dark/light defaults)


Pat9496:
Added link color (hyperlink text color) customization as requested — both highlight and link colors now live in the same View → Colors… window, with independent color pickers and reset buttons. See the latest commit for details.

aydiler:
Merged — thanks, and welcome as a first-time contributor here.

Verified locally against the merge result rather than the branch alone: link colour and highlight colour both apply from persisted state (checked with `highlight_color:Some((255,80,0,255))`, `link_color:Some((0,255,120,255))` — links go green in prose and inside tables, the active tab picks up the highlight), and with no override set the rendering is unchanged from `main`. All test suites pass, `cargo fmt` is clean.

Two things I liked: the `last_applied_*` mirroring so Visuals only rebuild when a colour actually changes, and that you registered the dialog widgets for MCP (`Colors Dialog: …`) per `.claude/rules/mcp-testing.md` — that is easy to miss.

One note for next time, no action needed now: the branch history contains `Delete docs/LESSONS.md` followed by `Restore docs/LESSONS.md, accidentally deleted in full`, and the restored copy was 26 lines shorter than `main` because `main` had moved on. I did a trial merge before merging for real to check — git resolves the delete-and-restore as a no-op against the merge base, so nothing was lost and the file is intact on `main`. Worth knowing that a "restore" commit restores the state at your branch point, not the current state of the target branch.

PR #133, opened 2026-09-04
Release v0.2.0

aydiler:
Prepares the 0.2.0 release. **Does not tag** — pushing `v0.2.0` publishes to crates.io, the Snap Store, AUR twice and GitHub Releases, and none of that is undone by deleting the tag.

## Why minor, not patch

121 commits and 39 merged PRs since v0.1.17 (22 August), most of them from @RichardCao and @Pat9496. Viewport-clipped rendering is back on for long documents, tables were rebuilt around content-driven column widths, and font fallback asks the platform instead of matching hardcoded family names. The rendered output of an unchanged document looks different — that is more than a patch.

## Contents

| File | Change |
|---|---|
| `CHANGELOG.md` | 0.2.0 entry; 38 of the 39 PRs described (only the test-only #82 is omitted) |
| `crates/egui_commonmark/CHANGELOG.md` | 0.28.0 section, renderer-scoped |
| `Cargo.toml` | md-viewer 0.1.17 → 0.2.0; fork refs 0.27.0 → 0.28.0 |
| `crates/egui_commonmark/Cargo.toml` | fork workspace 0.27.0 → 0.28.0 |
| both `Cargo.lock` | six name-anchored lines |
| `snap/snapcraft.yaml` | 0.2.0 (the `version-sync` job compares this against `Cargo.toml`) |
| `README.md` | Full Width, Formula Size and YAML frontmatter documented (separate commit) |

## Two findings worth recording

**The fork goes to 0.28.0, not 0.27.0.** crates.io holds 0.25.0 for all three fork crates, and md-viewer itself is at 0.1.15 — the local 0.26.0 and 0.27.0 bumps belong to the v0.1.16 and v0.1.17 releases whose crates.io publish never completed. 0.27.0 has been untouched since the last tag while the renderer changed substantially, so reusing it would ship very different code under a number two earlier release attempts already claimed. `scripts/check-fork-publishable.sh` now reports all three as fresh at 0.28.0.

**The lockfile trap from LESSONS.md is real here.** The fork lockfile carries `glam` in sixteen versions, including both `0.27.0` **and** `0.28.0`. A `sed` on the version line would have rewritten glam onto a version that already exists in the same file. Both lockfiles were edited by anchoring on `name = "…"` instead; the diff is exactly the six intended lines.

## Verification on this tree

- root: 60 passed, 1 ignored
- renderer: 65 lib + 18 wrapping + 1 integration + 16 doctests passed
- backend: 27 + 1 passed
- `cargo fmt --all -- --check`: clean
- `scripts/check-fork-publishable.sh`: passes

An initial renderer run reported seven doctest linker failures. That was disk pressure — ~8 GB of debug binaries left in `/tmp` from earlier work in the session — not a defect. It is the same signature already recorded in LESSONS.md; after freeing the space the identical command passes.

## After merge

`git tag v0.2.0 && git push origin v0.2.0`. `scripts/publish-crates.sh` publishes the fork crates in dependency order before md-viewer, so no manual crates.io step is needed.


aydiler:
Pushed one more commit: `39f7ab2`, the release checklist and the AppStream metadata.

Walking `PUBLISHING.md` for this release surfaced three defects in the checklist itself, one of which it had been silently failing since v0.1.3:

- **The AppStream release history stops at 0.1.3 (May 2026).** The checklist has always said "append a `<release>` entry to `data/io.github.aydiler.md-viewer.metainfo.xml`", and every release since has skipped it. That file is what GNOME Software and KDE Discover show for the Flatpak, so the listing carries no history at all. 0.2.0 is added. The 0.1.4–0.1.17 gap is deliberately not reconstructed — the per-version summaries no longer exist anywhere, and inventing them would be worse than the gap.
- **The checklist still said `git-cliff -o CHANGELOG.md`.** #132 removed that from `.claude/rules/release-workflow.md`, but the same instruction sat in `PUBLISHING.md` — which is where the checklist actually lives. I had only fixed half of it.
- **The expected release output was wrong**: 4 binaries including macOS x86_64, against a matrix with 3 targets. Intel Mac was dropped after its runner pool queued 24+ minutes on v0.1.3.

Added from this release's own mistakes: bump the vendored fork's workspace version when the renderer changed (the `Vendored fork is publishable to crates.io` step fails otherwise), and edit `Cargo.lock` anchored on the package name rather than the version line.

While validating the metainfo, `appstreamcli` also flagged a screenshot URL that 404s — `syntax-highlighting-light.png` has never existed in this repository. Replaced with two that do (math rendering, search), which also show features the listing never advertised. `appstreamcli validate` now passes cleanly, which matters because the Flathub linter runs the same check.

The three CI checks were green on `d9271d5`; this commit touches only Markdown and XML, so they should stay green.

PR #136, opened 2026-09-04
Gate math cache-key tests behind the math feature

RichardCao:
## Summary

- gate the two remaining `math_cache_hash` tests behind the `math` feature, matching the function they exercise
- exercise the backend package with `--no-default-features` in the existing renderer CI job
- leave production code and runtime behavior unchanged

## Why

`math_cache_hash` is compiled only when the `math` feature is enabled. Two cache-key tests were not gated, so the backend package failed to compile in its feature-minimal configuration even though the runtime library supports that configuration.

This PR is deliberately scoped to the backend package. It does not claim that every target in the complete vendored renderer workspace supports `--no-default-features`.

## Testing

- `cargo test --manifest-path crates/egui_commonmark/Cargo.toml -p egui_commonmark_backend_extended --no-default-features` (13 tests and 1 doctest passed)
- `cargo test --manifest-path crates/egui_commonmark/Cargo.toml -p egui_commonmark_backend_extended --all-features` (27 tests and 1 doctest passed)
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,7 @@ jobs:
       - name: Test renderer crates
         run: |
           cargo test --manifest-path crates/egui_commonmark/Cargo.toml -p egui_commonmark_extended --features "$RENDERER_FEATURES"
+          cargo test --manifest-path crates/egui_commonmark/Cargo.toml -p egui_commonmark_backend_extended --no-default-features
           cargo test --manifest-path crates/egui_commonmark/Cargo.toml -p egui_commonmark_backend_extended --all-features
           cargo test --manifest-path crates/egui_commonmark/Cargo.toml -p egui_commonmark_macros_extended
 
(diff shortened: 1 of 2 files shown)


aydiler:
Verified both directions before merging, because a misplaced `cfg` gate would have silently disabled the tests and left the suite green either way.

| | |
|---|---|
| `main`, `-p egui_commonmark_backend_extended --no-default-features` | **2 × `E0425: cannot find function math_cache_hash`** |
| this branch, same command | 13 tests pass |
| this branch, `--features math` | 21 tests pass |

And the three cache-key tests still run by name under `math`, rather than vanishing:

```
test misc::tests::math_cache_key_includes_exact_colors ... ok
test misc::tests::math_cache_key_includes_the_point_size ... ok
test misc::tests::math_cache_key_absorbs_sub_hundredth_size_jitter ... ok
```

Your premise reproduces exactly as described. Merging.

Issue #122, opened 2026-09-03
Discuss unifying the app and vendored renderer Cargo workspaces


RichardCao:
## Context

The application root and `crates/egui_commonmark` are currently separate Cargo workspaces with separate `Cargo.lock` files. The root then connects the vendored renderer through `[patch.crates-io]`.

This layout has useful properties: the three renderer crates can be tested and published as a standalone workspace, while the application can depend on their crates.io versions for publish verification. However, it also creates synchronization risks:

- a renderer dependency or feature can be changed in one workspace but not exercised by the other;
- the root dependency and patch versions can drift from the renderer workspace version;
- contributors must know which lockfile and which Cargo invocation applies;
- a patch can exist but be ineffective or unused without an obvious local failure.

CI now exercises both workspaces, which reduces the immediate testing gap, but it does not answer whether two workspaces are still the best long-term structure.

### A. One repository-level workspace

Make the application and the three renderer crates members of one workspace and use path/workspace dependencies for development.

Potential benefits:

- one dependency graph and lockfile;
- simpler local commands and cache behavior;
- fewer opportunities for version or feature drift.

Questions/risks:

- whether independent publication of the renderer crates and application stays straightforward;
- how `cargo publish` verification should test the registry dependency path rather than silently using local members;
- whether distribution tooling currently relies on the existing root lockfile/layout.

### B. Keep the two-workspace boundary and add synchronization checks

Preserve the current publication boundary, but add a small CI/script check that verifies at least:

- the root `egui_commonmark_extended` dependency and both root patch versions match `[workspace.package].version` in `crates/egui_commonmark/Cargo.toml`;
- all patch paths exist and resolve to the expected package names;
- both lockfiles are reproducible/up to date;
- the root build actually resolves the patched renderer packages during source builds;
- the release/publish checks continue to exercise the registry-compatible path.

## Suggested next step

Keep the current layout until the publication and distribution consequences of option A are demonstrated in a small experiment. In the short term, option B is low risk and makes the existing contract explicit. If a unified-workspace prototype simplifies both normal CI and crates.io publication without special-case rewrites, record that decision in an ADR before migrating.

aydiler:
Adding a data point rather than a preference, because I ran into exactly this friction today while working through the table and viewport PRs.

## What the split cost, concretely

**Running the renderer tests requires knowing an incantation.** The obvious command fails:

```
$ cargo test -p egui_commonmark_extended
[...]

PR #137, opened 2026-09-04
CI: validate app and renderer workspace synchronization

RichardCao:
## Summary

- validate the dependency contract between the application workspace and the vendored renderer workspace
- require both lockfiles to remain valid with `--locked`
- keep the renderer dependency and patch versions aligned with the renderer workspace version
- verify both renderer packages resolve to their expected vendored paths in the application dependency graph
- fail when Cargo reports an unused patch

## Scope

This is the short-term guard discussed in #122 while the repository keeps its two-workspace publication structure. It does not decide or close the longer-term workspace-unification discussion, and it does not change runtime behavior.

## Testing

- `bash -n scripts/check-workspace-sync.sh`
- `bash scripts/check-workspace-sync.sh`
- verified both root and renderer `cargo metadata --locked` paths
- verified both vendored packages in `cargo tree --locked -p md-viewer -e normal`
- fault-injected an extra unused local patch and confirmed the check fails
- `git diff --check`

Related: #122

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,9 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Check app/renderer workspace dependency contract
+        run: bash scripts/check-workspace-sync.sh
       - name: Check versions match
         run: |
           # Robust extraction tolerating quotes and spacing
--- a/scripts/check-workspace-sync.sh
+++ b/scripts/check-workspace-sync.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+repo_root=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
+cd "$repo_root"
+
+renderer_manifest="crates/egui_commonmark/Cargo.toml"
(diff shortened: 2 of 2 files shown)


aydiler:
Verified with negative controls rather than only a green run — a guard that has quietly stopped detecting looks exactly like a healthy repository.

Deliberately desynchronized the contract three ways and measured the script's own exit code (not a pipeline's):

| state | rc | output |
|---|---|---|
| healthy | **0** | `workspace dependency contract is synchronized at renderer version 0.28.0` |
| renderer workspace version bumped alone | **1** | `error: Cargo.toml pins egui_commonmark_extended 0.28.0; renderer workspace is 0.29.0` |
| root dependency version wrong | **1** | `error: Cargo.toml pins … 0.27.0; renderer workspace is 0.28.0` |
| `[patch.crates-io]` version wrong | **1** | same shape |
| restored | **0** | healthy again |

All three are exactly the mistake this guard exists for, and it catches each one.

Worth saying: this would have caught a real slip today. Preparing 0.2.0 I bumped the fork from 0.27.0 to 0.28.0 across five places by hand — the workspace manifest, two workspace dependency entries, the root dependency, and both `[patch.crates-io]` lines. Missing one is easy and the failure it produces afterwards is unrelated-looking.

Scope is right too: it guards the contract without pre-empting #122. Merging.

PR #138, opened 2026-09-04
CI: check macOS and Windows builds on pull requests

RichardCao:
## Summary

- add native `cargo check --locked` jobs for macOS and Windows
- keep the existing Linux job as the full lint/build/test path
- use a separate workflow so the lightweight portability guard remains independent of Linux CI

## Scope

These jobs validate that the application's locked production dependency graph compiles on the native GitHub-hosted macOS and Windows runners. They do not claim to validate linking, runtime behavior, packaging, or release artifacts; those remain covered by the release workflow and platform-specific testing.

## Local validation

- workflow YAML parsed successfully
- `cargo check --locked` passed on Linux as a syntax/dependency sanity check
- `git diff --check`

The actual macOS and Windows result is intentionally left to the native GitHub Actions jobs introduced by this PR.

### DIFF
--- a/.github/workflows/cross-platform.yml
+++ b/.github/workflows/cross-platform.yml
@@ -1,0 +1,27 @@
+name: Cross-platform check
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  check:
+    name: cargo check (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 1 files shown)


aydiler:
This closes a gap that has already cost a botched release, which is worth recording on the PR.

`ci.yml` is `ubuntu-latest` in all three jobs — there is no cross-platform coverage on pull requests today. The first Windows failure therefore surfaces at tag time, in the release workflow, when it is most expensive.

That is not hypothetical. The v0.1.16 release run [32581040315](https://github.com/aydiler/md-viewer/actions/runs/32581040315) failed with:

```
error[E0433]: cannot find `unix` in `os`
```

in `Build (windows-x86_64)`, which is why v0.1.16 needed three release runs. The fix landed afterwards as #70, gating the `UnixStream` import behind `cfg(unix)` — a one-line problem that a `cargo check` on `windows-latest` would have caught in the PR that introduced it.

Both new jobs ran green on this PR itself (macOS 2m42s, Windows 5m19s), so the workflow is demonstrated rather than assumed. `fail-fast: false`, a 30-minute cap and `--locked` are all right, and the scope note is accurate — this checks the application's locked dependency graph compiles, not linking, runtime or packaging.

Merging.

Issue #141, opened 2026-09-04
Define consistent local path and file URI semantics


RichardCao:
## Summary

Define one explicit, cross-platform contract for converting between Markdown link destinations, `file://` URIs, and filesystem paths before consolidating the current path-handling code.

This is a design/bug-hardening issue, not a proposal for a one-shot refactor. Existing local Markdown navigation should remain unchanged until edge-case behavior is agreed and covered by tests.

## Current responsibilities

The application currently handles related operations in several places:

- file and folder picker byte output;
- tab opening, canonicalization, deduplication, and history;
- parsing explicit Markdown links and bare `.md` references;
- registering renderer link hooks;
- resolving relative destinations against the current document directory;
- decoding percent-encoded paths and converting `file://` URLs;
- copying filesystem paths back as `file://` text;
- watcher and Explorer path comparisons.

Each path is reasonable in isolation, but the combined behavior is not represented by one documented type or policy.

## Semantics to decide

1. **Relative paths:** always resolve against the containing document, including links opened from history or a reloaded tab.
2. **Fragments and queries:** define whether `guide.md#section` opens the document only or also navigates to the heading; avoid treating `#fragment` as a filesystem path.
3. **Percent encoding:** decode URL path components exactly once; preserve literal percent signs and invalid escape sequences predictably.
4. **Platform paths:** cover Unix absolute paths, Windows drive-letter paths, UNC paths, and `file://host/...` authority rules.
5. **Non-UTF-8 paths:** preserve picker/Explorer paths on Unix without forcing lossy UTF-8; define what can and cannot be represented as a Markdown URL.
6. **Missing targets:** decide whether unresolved links remain external-looking, show an error, or may open a tab that can recover when the file appears.
7. **Canonicalization:** avoid requiring the target to exist merely to normalize identity; define symlink behavior and tab deduplication.
8. **Security boundary:** local navigation must not reinterpret `http`, `https`, `data`, or unknown schemes as filesystem paths.

## Proposed incremental shape

- introduce table-driven tests for the agreed destination-to-target behavior on each platform;
- extract a small pure resolver returning a typed result such as document fragment, local path, or external URL;
- migrate explicit Markdown links first;
- migrate bare-path auto-link registration and copied file URIs separately;
- only then consolidate tab/history/watcher identity if the same normalization contract applies.

Each implementation PR should preserve current behavior outside the cases it explicitly fixes and should not be coupled to the larger `main.rs` decomposition.

[...]

PR #154, opened 2026-09-05
Fix local-link URI scheme classification

RichardCao:
## Summary

- recognize explicit URI schemes using the RFC 3986 scheme grammar
- register and resolve only scheme-less paths and `file:` URIs as local filesystem navigation
- keep Windows one-letter drive prefixes classified as filesystem paths
- add regressions for `data:`, `javascript:`, unknown schemes, local paths, file URIs, and Windows drive paths

Non-`file:` schemes are no longer interpreted or registered as local filesystem links; they continue through the existing external-link handling. This PR does not claim to disable those schemes globally or address an execution vulnerability.

This is the first bounded fix discussed in #141. Fragment-only links, missing-target behavior, canonicalization, non-UTF-8 paths, and broader Windows/UNC policy remain outside this PR.

## Testing

- `cargo test --locked` (64 passed, 1 ignored)
- `cargo clippy --locked --all-targets -- -D warnings`
- `git diff --check`

Closes one part of #141.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1240,17 +1240,14 @@ fn register_existing_markdown_path(
 
 /// Check if a link destination points to a local markdown file
 fn is_local_markdown_link(destination: &str) -> bool {
-    if destination.starts_with("http://")
-        || destination.starts_with("https://")
-        || destination.starts_with("mailto:")
-        || destination.starts_with("tel:")
-        || destination.starts_with("ftp://")
-        || destination.starts_with('#')
-    {
+    if destination.starts_with('#') {
         return false;
     }
 
-    let path = if destination.starts_with("file://") {
+    let path = if let Some(scheme) = explicit_uri_scheme(destination) {
+        if !scheme.eq_ignore_ascii_case("file") {
+            return false;
+        }
         let Ok(url) = url::Url::parse(destination) else {
             return false;
         };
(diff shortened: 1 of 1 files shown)


aydiler:
Verified against the baseline I recorded on #141, using my own case list rather than your tests — so this is an independent check, not a re-run of yours.

| destination | #141 baseline | this PR |
|---|---|---|
| `data:text/plain,x.md` | **local = true** | not local ✓ |
| `javascript:void(0)/x.md` | **local = true** | not local ✓ |
| `unknownscheme://host/a.md` | **local = true** | not local ✓ |
| `http://…`, `mailto:…`, `#anchor` | not local | not local ✓ |
| `guide.md`, `/abs/path/doc.md` | local | local ✓ |
| `docs%20with%20spaces/guide.md` | local | local ✓ |
| `100%done.md` (literal `%`, invalid escape) | local | local ✓ |
| `file:///tmp/x.md` | local | local ✓ |

All three destinations that were wrongly classified are fixed, and every positive control from #147 survives.

**One false alarm of mine, worth recording rather than hiding:** my case list also asserted `C:/Users/x/doc.md` should stay local, and it failed. That was my error, not yours — the drive-letter exception is `#[cfg(target_os = "windows")]`, and on Linux `C` genuinely *is* a scheme under RFC 3986. I had written a platform-independent expectation for platform-dependent behaviour. Your implementation is right.

## One real gap, and it is not in this PR

`windows_drive_prefixes_are_not_treated_as_uri_schemes` is the only assertion covering the drive-letter branch, and **CI never runs it.** The cross-platform workflow from #138 runs `cargo check --locked`, which does not build test targets — so on Windows that test is not merely unexecuted, it is not even compiled. The `#[cfg]` branch it guards is the one piece of this PR that cannot be exercised on Linux.

Not a reason to hold this. But the safety net around the branch is thinner than it looks, and I added that workflow, so the gap is mine. Two options if you think it is worth closing: add `--all-targets` to the cross-platform check so the code at least compiles there, or promote it to `cargo test` on Windows. I am happy to open either as a follow-up — your call, since it is your branch that would be relying on it.

Local verification of this PR: application 64 passed / 1 ignored, `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean. The scope note in your description matches what the diff does — fragments, missing-target feedback, canonicalisation, non-UTF-8 and broader Windows/UNC policy are all untouched.

Issue #118, opened 2026-08-31
Optimize table column widths for lower wrapped height


RichardCao:
## Context

Follow-up to #112 / #115 and the row-height fixes in #116.

The fitted-column policy in #115 deliberately optimizes fairness and relative natural-width demand. It is stable and continuous, but it does not ask whether giving a column another few pixels would remove one or more wrapped lines.

This is visible in two common layouts:

- a two-column schema table with short field names and long prose requirements;
- a dense report table with many numeric/status columns plus one identifier, explanation, or report-links column.

In the latter case, the hard minimum alone can consume the viewport. For example, 13 columns at 40 px require 520 px before separators and the outer frame. When the visible budget is near or below that total, every column receives 40 px and the long-content column becomes unnecessarily tall. A different fixed-width allocation has little or no room to improve it; reducing height may require bounded horizontal overflow.

## Desired behavior

- prefer allocations that reduce the total rendered table height, not only unwrapped natural-width error;
- retain the #112 invariants where applicable: deterministic output, hard minimums, natural-width caps, and stable behavior under small width changes;
- preserve manual column resizing while the table bound is unchanged;
- recompute only when the table layout bound/content/font metrics change, not every frame;
- retain horizontal scrolling as the fallback for dense tables;
- avoid column-type or font-family name heuristics such as hard-coded “numeric”, “identifier”, or “prose” widths.

### Phase 1: height-aware fitting inside the existing budget

1. Keep `fit_column_widths` from #115 as the deterministic baseline.
2. Reuse the production `table_cell_height` measurement to build quantized height curves for each cell/column (an 8 px width quantum should be sufficient initially).
3. Score a candidate allocation primarily by:

   `sum(row cells.map(height).max())`

   This matches the height TableBuilder actually assigns to heterogeneous rows. Use total cell height as a secondary tie-breaker so the search can cross flat regions where a different column temporarily remains the row maximum.
4. Starting from the #115 allocation, perform bounded coordinate descent: transfer one width quantum from a donor column to a receiver only when the score improves, while preserving the total budget and all bounds.
5. Use stable column-index tie-breaking and a strict iteration cap. Cache the resulting initial widths under the same layout signature/bound lifecycle used by the resizable table state.

This approach measures the real wrapping behavior, including mixed proportional text, monospace inline code, links, configured line height, and conservative math height. It avoids guessing content types.

### Phase 2: dense-table overflow policy

[...]

PR #157, opened 2026-09-05
Optimize fitted table widths for lower wrapped height

RichardCao:
## Summary

- keep the existing deterministic allocator and per-column Unicode word floors as the baseline
- refine only the initial fitted widths with a bounded, deterministic search that measures production row heights
- cache optimized widths by content, bound, font/typography, pixel scale, layout revision, and math scale
- preserve manually resized TableBuilder widths while that layout key remains stable
- apply the same policy to Markdown and HTML tables

This is the fixed-width Phase 1 discussed in #118. It deliberately does not compress header word floors or introduce horizontal overflow. Dense tables where every column is already at its floor remain a separate Phase 2 policy decision.

## Bounds and behavior

- searches every 8 px transfer through a bounded 64 px look-ahead window, for at most eight passes
- measures at most 4096 cells per optimization
- accepts row-height improvements deterministically and permits at most one speculative secondary-score move to cross a two-way height plateau
- rolls back to the last primary improvement if the next move cannot reduce total row maxima or the measurement budget is exhausted
- resets cached initial widths only when the layout key changes

## Real report check

I tested the 13-column family table from the motivating report using all 15 parsed rows (header included), measuring each real cell separately:

| Width budget | Sum of row maxima, baseline -> optimized | Measured cells |
|---:|---:|---:|
| 900 px | 1643.7 -> 670.7 | 1665 |
| 1100 px | 685.7 -> 655.7 | 1785 |
| 1400 px | 655.7 -> 461.1 | 1755 |

In an unoptimized debug build with font/layout caches warmed, the cached initial calculation took about 49 ms, a stable-key hit took about 11 microseconds with zero cell remeasurement, and a width-change recomputation took about 53 ms. These timings are diagnostic rather than a release benchmark; the hard measurement count is the enforced performance bound.

The new Markdown and HTML production geometry regressions were also run against clean `main`: both fail there and pass with this change.

## Testing

- `cargo test --locked` (root: 64 passed, 1 ignored)
- renderer workspace with production features (90 library, 1 slice perf, 25 wrapping, macros/trybuild and doc tests passed)
- backend `--no-default-features` (14 + doctest passed)
- backend `--all-features` (28 + doctest passed)
- `cargo clippy --locked --all-targets -- -D warnings`

Relates to #118.

### DIFF
--- a/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
@@ -1,3 +1,6 @@
+use std::collections::HashMap;
+use std::fmt::Write;
+use std::hash::{Hash, Hasher};
 use std::iter::Peekable;
 use std::ops::Range;
 
@@ -472,6 +475,7 @@ fn table_cell_height(
         height += inline_math_height
             + (ui.spacing().item_spacing.y + 1.0) * inline_math_count as f32;
     }
+
     height
 }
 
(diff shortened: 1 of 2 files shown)


aydiler:
Verified this end to end, including visually. It does what it says, and the numbers hold up.

## It works, and the result is visibly better

Six columns of competing prose at a 1500 px window, `main` against this branch:

| | main | this PR |
|---|---|---|
| table bottom | ~647 px | **~573 px** |
| first row | 5 lines | **4 lines** |
| rows 2 and 4 | 4 lines each | **3 lines each** |
| differing pixels | — | 116 353, bbox confined to the table |

The width moves to `Requirement` and `Rationale`, which is exactly where the wraps were.

## It took me three failed attempts to see it, and that is worth telling you

My first three fixtures showed **zero** differing pixels, and I nearly wrote "does not engage". All three were in the region this PR deliberately excludes. Instrumenting `optimize_fitted_widths` showed why:

```
OPT n=6 can_transfer=false
  baseline=[104.78, 113.78, 86.03, 66.59, 92.88, 59.66]
       min=[104.78, 113.78, 86.03, 66.59, 92.88, 59.66]
```

`baseline == minimums` exactly — every column already at its word floor, so no donor has slack and `can_transfer` is false. That is your Phase 2 dense case, correctly untouched.

What put me there was the **side panels**: at a 620 px window the explorer and outline leave the document pane far too narrow, so everything collapses to floors. Phase 1 needs the pane between the floor sum (~524 px here) and the desired sum (~1292 px). Not a defect — but it means the improvement is invisible on a narrow window, which is where a user is most likely to want it. Worth a sentence in the issue when Phase 2 is scoped.

## Resize cost

The layout key includes the bound, so a window drag misses the cache on every step. I measured CPU jiffies across 24 resizes that all re-optimise:

| | CPU over 24 resizes |
|---|---|
| main | 109 jiffies |
| this PR | **110 jiffies** |

Essentially free on a 6×5 table. That does not contradict your ~53 ms — mine is 6×5, yours is 13×15 — but it does mean the cost scales with table size rather than being a fixed per-resize charge, and the measurement cap is what bounds it. On a very large table a continuous drag would be the case to watch.

## Guards

Neither runs in CI, so I ran both:

| | verdict |
|---|---|
| `scripts/scroll-regression.sh` | **PASS**, bottom at frame 68 |
| `scripts/visual-regression.sh` | **PASS**, no drift |

Frame 68 against main's 85 is the same effect from the other side: the document is shorter.

## Two notes, neither blocking

- **Rebase would help.** Your base `f65ee11` predates #158, so the guard scripts on this branch still do a blanket `pkill -9 Xvfb` — running them here restarted this machine's `xvfb99.service` twice. Fixed on main.
- **Pre-existing, not yours:** `Owner` and `Mitigation` touch without a gap (`renderingclamp the bias`) whenever a cell exactly fills its column. Present identically on main; mentioning it so it does not get attributed to this change later.

CI is five-way green, and the macOS/Windows jobs now compile test targets thanks to #155 — worth having on a PR this size.

PR #164, opened 2026-09-05
Keep visible spacing between Markdown table cells

### LINKED ISSUE #163: Keep visible spacing between fitted Markdown table cells
## Summary

Adjacent Markdown table cells can render with almost no visible gap when wrapped text exactly fills a fitted column. This was independently noticed while visually verifying #157, but it is present on `main` without #157-specific behavior.

## Reproduction on current main

Tested at `8a6ef8c` with:

```markdown
| Owner | Mitigation |
|---|---|
| preserve the final visual rendering | clamp the bias during processing |
```

A production `CommonMarkViewer` two-pass geometry probe measured:

| document width | gap between the two painted text rects |
|---:|---:|
| 308 px | 0.94 px |
| 412 px | 1.34 px |

This produces the visible `renderingclamp`-style join noted during #157 review.

## Root cause

The document renderer intentionally sets `ui.spacing_mut().item_spacing.x = 0.0` for inline Markdown flow. The nested `TableBuilder` inherits that value, so its cells receive no inter-column spacing. HTML table cells already add an explicit `Margin::symmetric(8, 4)`, but Markdown table cells render their `horizontal_wrapped` contents directly with no cell margin.

The existing Markdown height estimator already uses `(column_width - 8.0)`, so its wrapping geometry models 8 px of total horizontal inset even though the production Markdown cell currently does not apply that inset. Its separate vertical allowance is already needed by mixed-content row-height behavior and should not be turned into visible top/bottom padding.

## Proposed bounded fix

- wrap only Markdown cell contents in a `Frame::NONE` with `Margin::symmetric(4, 0)`;
- pass the resulting inner width to the existing recursive cell renderer;
- leave TableBuilder spacing, fitted-width allocation, HTML rendering, and the outer table width contract unchanged;
- add a production geometry regression asserting a useful gap between adjacent fitted cells, plus retain the existing wrapping/row-height regressions.

This is deliberately a cell-content padding fix, not another column-allocation change.

RichardCao:
Closes #163.

## Summary

- give Markdown table cells 4 px of horizontal inset on each side
- render cell contents against the resulting inner width, matching the existing height estimators’ 8 px allowance
- keep vertical inset at zero so fitted row heights and clipping behavior do not change

This does not change the table width allocator, column bounds, divider positions, manual resize state, or the existing HTML-table path.

On current `main`, the production geometry fixture left only about 0.94 px between adjacent cells at a 308 px document width (1.34 px at 412 px). The change preserves an 8 px visible gap.

## Testing

- root test suite: 64 passed, 1 ignored
- renderer library with features: 90 passed
- wrapping tests: 26 passed
- slice performance test passed
- macros and trybuild tests passed
- documentation tests: 16 passed, 1 ignored
- `git diff --check`

The exact commit was reviewed at high reasoning depth after the full test run; no blocking issues were found.

### DIFF
--- a/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
@@ -2224,36 +2224,49 @@ impl CommonMarkViewerInternal {
                                                     row_ui.col(|ui| {
                                                         ui.style_mut().wrap_mode =
                                                             Some(egui::TextWrapMode::Wrap);
-                                                        let col_w = ui.max_rect().width();
-                                                        ui.set_width(col_w);
-                                                        // Isolate the wrapping cursor from the
-                                                        // preallocated TableBuilder row height.
-                                                        // Otherwise egui uses that entire height as
-                                                        // the first text line's minimum height and
-                                                        // later inline widgets overflow the row.
-                                                        ui.horizontal_wrapped(|ui| {
-                                                            for (e, src_span) in col {
-                                                                let tmp_start = std::mem::replace(
-                                                                    &mut self.line.should_start_newline,
-                                                                    false,
-                                                                );
-                                                                let tmp_end = std::mem::replace(
-                                                                    &mut self.line.should_end_newline,
-                                                                    false,
-                                                                );
-                                                                self.event(
-                                                                    ui,
-                                                                    e.clone(),
-                                                                    src_span.clone(),
-                                                                    cache,
-                                                                    options,
-                                                                    col_w,
-                                                                );
-                                                                self.line.should_start_newline =
-                                                                    tmp_start;
-                                                                self.line.should_end_newline = tmp_end;
-                                                            }
-                                                        });
+                                                        ui.set_width(ui.max_rect().width());
+                                                        egui::Frame::NONE
+                                                            .inner_margin(egui::Margin::symmetric(4, 0))
+                                                            .show(ui, |ui| {
+                                                                let col_w = ui.available_width();
+                                                                ui.set_width(col_w);
+                                                                // Isolate the wrapping cursor from the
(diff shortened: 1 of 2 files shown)


aydiler:
Verified on your branch. **Everything passes except one guard, and that failure is the guard's fault, not yours.**

| | |
|---|---|
| application tests | 64 passed, 1 ignored |
| renderer | 90 lib + 1 slice + 26 wrapping + 16 doctests |
| `scripts/scroll-regression.sh` | **PASS** |
| `scripts/visual-regression.sh` | **FAIL** — see below |

## The visual guard will go red on this change, by design

```
/tmp/shot-visreg-1.png: content starts at x=235, baseline is x=230 (5px shift)
/tmp/shot-visreg-3.png: content starts at x=236, baseline is x=230 (6px shift)
```

That guard keys off the leftmost painted pixel. This PR adds exactly 4 px of inset per cell side — so table content **should** start further right, and the guard reports the intended change as drift. `docs/LESSONS.md` already records this weakness: *"keys off the leftmost painted pixel, which works on its own flush-left fixture but false-positives on any document with indented content"*.

So this is not a reason to hold the PR. But it does mean **`visual-regression.sh` will stay red after merge** until its baseline is regenerated or its metric replaced. Worth deciding which — I am happy to do either as a follow-up, and it is arguably my mess since I have been leaning on that guard all day.

## On the fix itself

Making the paint match the estimator is the right direction rather than the reverse. The estimator's `column_width - 8.0` was the intended design; the paint never honoured it. That is the same shape as the `parse_row` boundary bug — measurement and paint disagreeing about a cell — and resolving it toward the *measurement* keeps the height reservation from #131 and the optimizer from #157 consistent with what is actually drawn.

I also checked how old the underlying gap is and posted it on #163: it reproduces identically on a `v0.1.17` build, so it predates the entire 0.2.0 table effort. Fitted columns did not cause it, they only made it common.

Issue #167, opened 2026-09-06
Frontmatter document paints nothing below the first paragraph at narrow window widths


aydiler:
## Symptom

A document with a YAML frontmatter block, opened in a window **1040 px wide**, paints the frontmatter table, the `h1` and the first paragraph — and then nothing. Roughly 360 px of the document pane stays blank, even though:

- the outline sidebar lists a later heading (`Fit summary`), so the document *is* parsed, and
- the document pane shows a scroll thumb covering only ~40% of its track, so the renderer knows there is more content.

Widening the window to 1080 px makes the whole document appear.

## Reproduction

Fixture (`dark-energy-fit.md`):

```markdown
---
title: Quintom dark energy — DESI DR2 fit
authors: A. Diler, R. Cao
date: 2026-09-05
status: draft
tags: cosmology, dark-energy, DESI
reviewers:
  - platform
  - rendering
abstract: >
  A two-field quintom model fitted against the DESI DR2 BAO
  compilation, with residuals marginalised over nuisance N.
---

# Quintom dark energy

The block above is YAML frontmatter. Instead of showing it as raw text or
dropping it, md-viewer renders it as a key/value table — the way VS Code does.

Nested mappings, sequence items and folded scalars are appended to the
preceding value verbatim, so nothing in the source is lost.

## Fit summary

| model | χ² (13 pts) | Ω_m | w_0 |
|---|---|---|---|
| ΛCDM | 10.28 | 0.297 | −1 |
| quintom (free fit) | 7.31 | 0.326 | −0.716 |
```

`md-viewer dark-energy-fit.md`, window resized to 1040x750.

## Measurements

Painted pixels in the document pane below y=380 (Xvfb, isolated `XDG_*`, `main` @ 685c382):

| window | painted px | content |
|---|---|---|
| 1040x750 | 2750 | **missing** |
| 1040x750 (repeat) | 2750 | **missing** |
| 1040x750 (repeat) | 2750 | **missing** |
| 1040x900 | 2938 | **missing** |
| 1080x750 | 11372 | complete |
| 1120x750 | 11372 | complete |
| 1160x750 | 11372 | complete |
| 1200x750 | 11372 | complete |
| 1400x750 | 11372 | complete |
| 1600x900 | 11368 | complete |

Fully deterministic — three independent runs at 1040 produce the identical figure. Height is not a factor (1040x900 fails too); width is.

## What narrows it to the frontmatter table

[...]

Issue #140, opened 2026-09-04
Intermittent blank frame when scroll state exceeds updated content extent


RichardCao:
## Summary

After the table-height fixes in #131, an intermittent one-frame blank document pane can still be observed while scrolling long documents. It has not yet been reduced to the table-boundary failure fixed by #131.

One distinct candidate path is a stored vertical scroll offset becoming temporarily larger than the document's updated content extent after asynchronous layout changes.

## Why this is separate from earlier fixes

- #113 made viewport slices reproduce bootstrap layout and fixed table/navigation slice identity.
- #114 fixed Full Width table bounds and a separate deep-scroll blank position.
- #131 fixed skipped tables reserving no height when the entire table was above the viewport.

Those fixes address incorrect content extent or slice layout. They do not explicitly clamp an already-persisted `egui::scroll_area::State::offset` before the next viewport slice is selected when content later becomes shorter.

## Candidate failure sequence

1. A long document is scrolled near its previous bottom.
2. An asynchronous image, formula, font, or table-layout update changes the measured content height.
3. The stored offset is now beyond `content_height - viewport_height`.
4. `show_viewport` receives an out-of-range viewport for one frame and the virtualized event range paints no visible document content.
5. Egui clamps the offset after painting, so the following frame recovers.

This matches an intermittent single blank frame, but remains a hypothesis until correlated with a real capture.

## Proposed regression

A renderer test can:

- bootstrap a long virtualized document;
- store an offset deliberately beyond its measured content height;
- render the next frame;
- assert that visible text is painted in the document clip rather than producing a blank frame.

The candidate implementation would clamp both persisted and requested offsets to the current maximum before viewport slice selection, while leaving egui's normal post-layout clamp in place.

## Sequencing

This touches `show_scrollable` and `tests/wrapping.rs`, the same areas being updated by #116. Discuss and collect reproduction evidence now, but prepare the code PR only after #116 lands so the fix and regression are reviewed against the final row-height behavior.

aydiler:
Your candidate path is visible in the code, and the mechanism is structural rather than incidental — the clamp exists, but it is **one frame late by construction**.

`show_scrollable`'s slice branch:

| line | what happens |
|---|---|
| 1229 | `make_scroll_area().show_viewport(...)` — the closure runs with the **stored** offset |
| 1256 / 1265 | `partition_point` selects the slice from `viewport.min.y` / `viewport.max.y`, i.e. from that same stored offset |
[...]

PR #176, opened 2026-09-07
diag: report how a viewport slice's event range is chosen (#140)

aydiler:
For #140. **Draft on purpose — do not merge before `v0.2.0` is tagged.**

## Why a second probe

`MDV_DIAG_SLICE` reports where a slice is *placed*. On #167 it did exactly its job and was **still not enough**: it reported zero off-screen placements in *both* the working and the broken run. That true negative correctly excluded the placement hypothesis — and could say nothing about what was actually wrong.

The cause was in how the event *range* gets chosen:

```rust
let below = split_points.partition_point(|(_, start, _)| start.y <= viewport.max.y);
let last_event_index = split_points.get(below + 1)...;
```

`below` had collapsed to 1, so the range ended at event 11 of 63 and everything below the frontmatter table went unpainted. A throwaway probe showed that in a single frame. #140 is still open and is the same class of failure, so this makes the instrument permanent instead of rebuilding it next time.

## What it prints

One line per frame:

```
DIAG split viewport=[876,1626] extent=1626 above=10 below=17 events=[32,134)/134 sp=17
```

`extent` is `page_size.y`, the height the bootstrap pass measured. **#140's stated candidate path is a stored scroll offset that briefly exceeds the updated content extent**, so those two values sit side by side and the line is marked `OFFSET>EXTENT` when the viewport runs past the extent.

The split-point table is dumped only when the selection looks degenerate — an empty range, or one stopping at the first boundary while the viewport plainly reaches further. Seventeen rows every frame would bury the signal on a long document.

**The summary line is unconditional while enabled**, so its silence means the probe was not running rather than that nothing happened. That distinction is why `diag_report_slice` is written the same way, and it is what made its null result on #167 trustworthy.

## Verification — both halves

Same document, same harness:

| run | DIAG lines |
|---|---|
| `MDV_DIAG_SPLIT=1` | **486** |
| without the variable | **0** |

On the healthy run, `OFFSET>EXTENT` and degenerate-table dumps were both **0** — genuine zeroes from a probe proven to be reporting, not an unimplemented check that would report zero either way. At the document bottom the line reads `viewport=[876,1626] extent=1626`: equal, never exceeding, which is the post-render clamp from `docs/LESSONS.md` behaving as documented.

## Cost when disabled

One relaxed bool load per painted slice, same `OnceLock` pattern as the existing probe.

## Release timing

Touches the vendored renderer (workspace 0.28.0, not yet published). Held until the tag so the release that has been verified end to end is not disturbed.

### DIFF
--- a/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
@@ -1102,6 +1102,67 @@ events={events}{}",
     );
 }
 
+/// Diagnostic instrument for issue #140, companion to [`diag_report_slice`].
+///
+/// Inert unless `MDV_DIAG_SPLIT` is set. Where `diag_report_slice` reports
+/// where a slice is *placed*, this reports how its event range was *chosen* —
+/// the split-point table and both `partition_point` results that select it.
+///
+/// #140's candidate path is a stored scroll offset that briefly exceeds the
+/// document's updated extent after an asynchronous layout change, so the two
+/// values that decide that are printed side by side: the viewport the frame
+/// shows, and `page_size.y`, the extent the bootstrap pass measured. When the
+/// former runs past the latter the line is marked `OFFSET>EXTENT`.
+///
+/// It earned its place on #167, where `MDV_DIAG_SLICE` correctly reported no
+/// off-screen placement in either the working or the broken run — a true
+/// negative that excluded the placement hypothesis but could not say what was
+/// wrong. This probe showed it in one frame: `below` had collapsed to 1, so
+/// the range ended at event 11 of 63 and everything below the frontmatter
+/// table went unpainted.
+///
+/// The summary line prints on **every** frame, so its silence means the probe
+/// was not running rather than that nothing happened. The full table is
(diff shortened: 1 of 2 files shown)


aydiler:
One more verification, closing the question this probe is most vulnerable to: **is the `extent` field actually tracking anything, or does it just happen to print a constant?**

It tracks. In the verification run on a document that was resized mid-session, `extent` took three distinct values:

```
extent=1740
extent=1622
extent=1626
```

That matters for how its output should be read. In the two #140 shrink attempts the extent was constant — 2 distinct values in one, exactly 1 in the other, across ~2000 frames each. Because the field is proven to move when layout moves, that constancy is a real property of those fixtures rather than a dead readout, which is what lets the `OFFSET>EXTENT` count of zero be interpreted at all: the condition never arose, rather than the probe being unable to see it.

Full verification matrix for this PR:

| question | answer | evidence |
|---|---|---|
| does it report when enabled? | yes | 486 lines |
| is it silent when disabled? | yes | 0 lines, same document |
| does `extent` track layout changes? | yes | 3 distinct values under resize |
| does it stay quiet on a healthy run? | yes | 0 `OFFSET>EXTENT`, 0 table dumps |

The fourth row is only meaningful because of the first three.

PR #181, opened 2026-09-08
Report missing local link targets

RichardCao:
Part of #141 (missing-target feedback only).

## Summary

- return canonicalization failures from the shared local-link resolver instead of converting them to a silent no-op
- route both plain click and Ctrl/Cmd+click through one result handler that feeds the existing error bar
- leave the current document, tab set, and navigation history untouched when the target is missing
- keep bare-filename discovery existence-gated, while explicit Markdown links can report a broken target

This deliberately does not change fragment handling, URI-scheme classification, non-UTF-8 behavior, canonical tab identity, symlink resolution, or the behavior of direct Explorer paths.

## Regression evidence

The new test calls the same production result handler used by `render_tab_content` and asserts the complete platform-native error message for both modifier paths, plus unchanged path/back/forward state.

I also restored the old swallowed-error behavior temporarily while keeping the test in place. It failed as intended:

```text
left: None
right: Some("Unable to open missing.md: No such file or directory (os error 2)")
```

Restoring the fix made it pass.

## Testing

- `cargo test --locked`: 65 passed, 1 environment-dependent font test ignored
- `cargo clippy --locked -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

The final commit was reviewed at high reasoning depth against current `main`; no blocking issues were found.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1047,26 +1047,13 @@ impl Tab {
     }
 
     fn navigate_to_link(&mut self, link: &str) -> io::Result<bool> {
-        if link.starts_with('#') {
-            return Ok(false);
-        }
-
-        let Some(current_dir) = self.path.parent() else {
-            return Ok(false);
-        };
-
-        // Share one resolver with `resolve_link` (the ctrl+click path).
-        // These used to differ: #78 taught percent-decoding and `file://`
-        // handling to `resolve_local_link_path`, but this function still did a
-        // raw `join`, so a link with an encoded space opened in a new tab on
+        // Share resolution and error handling with the ctrl+click path. These
+        // used to differ: #78 taught percent-decoding and `file://` handling
+        // to `resolve_local_link_path`, but this function still did a raw
+        // `join`, so a link with an encoded space opened in a new tab on
         // ctrl+click and silently did nothing on a plain click.
-        let Some(target_path) = resolve_local_link_path(link, current_dir) else {
+        let Some(target_path) = self.resolve_link(link)? else {
             return Ok(false);
-        };
-
-        let target_path = match target_path.canonicalize() {
-            Ok(p) => p,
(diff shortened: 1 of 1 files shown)


aydiler (approved):
Reviewed against the six points from #141 and verified rather than read. All six hold.

**The two swallow sites are both gone, and the duplication with them.** `navigate_to_link` now delegates to `resolve_link` instead of resolving a second time, so the #147 divergence class cannot recur here — that is more than I asked for and the right call.

**I ran your test red before trusting it.** Restoring the old swallow (`Ok(target_path.canonicalize().ok())`):

```
test missing_explicit_link_surfaces_errors_without_changing_tab_or_history ... FAILED
assertion `left == right` failed
  left: None
 right: Some("Unable to open missing.md: No such file or directory (os error 2)")
```

and green again on your version. It detects exactly the defect it targets.

Deriving `expected_error` from a real `canonicalize().unwrap_err()` rather than a hardcoded string is the detail I would have got wrong — the message differs by platform, and a literal would have made this a Linux-only test.

**Test inventory unchanged plus one**, compared by name rather than count against `main`:

```
in main, missing from PR:  (none)
new in PR:                 tests::missing_explicit_link_surfaces_errors_without_changing_tab_or_history
```

Root suite 65 pass / 1 ignored, all five CI jobs green.

Point 4 is satisfied structurally rather than by convention: resolution happens before any history or tab mutation, so an error cannot leave a half-navigated tab. Your test asserts both directions of history plus the tab path, for both modifier paths.

One note, not a change request: `handle_local_link_click` takes `&mut Tab` even on the ctrl+click path, where `resolve_link` only needs `&self`. Harmless, and the uniform signature is what lets both paths share one result handler — worth leaving as is.

Merging. Thanks for the clean boundary; point 7 stays open as its own design question, as you proposed.

PR #182, opened 2026-09-08
Optimize dense tables with bounded overflow

RichardCao:
Follow-up to #118.

## Summary

- activate bounded overflow only when header-word floors already exceed the visible column budget, so the feature never makes a fitting table start scrolling
- score the complete `0, 32, 64, 96, 128, 160` candidate set, clipped by a 30% floor-total cap and natural width, then choose the smallest candidate reaching 90% of the best row-max improvement
- reuse measured cell heights across candidates with a deterministic 4096-cell allowance per candidate and fail closed when a full baseline score cannot fit
- keep the horizontal viewport unchanged and preserve stable manual column widths
- reuse dense layout measurements while the viewport remains below the same floor total, avoiding expensive rescoring on every divider pixel

The 30% proportional cap is the value confirmed in #118: it still admits the measured +160 px knee for the 593.5 px 13-column floor total (`160 / 593.5 = 26.96%`), while scaling the cap down for narrower tables.

## Regression coverage

- in-repository 13-column synthetic knee fixture
- adversarial dip/recover scorer contract proving every candidate is visited (no early exit or binary search)
- 90% smallest-candidate selection
- absolute, proportional, natural-width, and exact terminal caps
- dense-only activation, no-improvement fallback, and oversized-table fail-closed behavior
- cache reuse within the dense range and invalidation across the floor threshold
- exact 199.6 px dense / 200.4 px fitting boundary regression, including the cached-width result
- production Markdown and HTML geometry tests; removing either renderer connection makes its test fail

## Validation

- rebased onto current `main` (`d5e3229`)
- `cargo test --locked` — 67 passed, 1 ignored
- renderer `cargo test ... --features math --lib` — 93 passed
- renderer wrapping tests with math — 30 passed
- backend tests with math — 22 passed plus 1 doc-test
- boundary test red-before-green: removing the branch discriminator reproduces the 199.6/200.4 key collision
- `git diff --check`
- high-depth independent review after the cache-key correction: no blockers

A local release-mode probe of the original 308-line report measured approximately 250 ms for first dense layout, then 65 ms and 46 ms for successive narrower viewport changes inside the same dense range, and 24 ms for a stable frame. The probe and its absolute local path were not committed.

### DIFF
--- a/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
+++ b/crates/egui_commonmark/egui_commonmark/src/parsers/pulldown.rs
@@ -791,10 +791,28 @@ fn optimize_fitted_widths(
     row_count: usize,
     mut measure_column: impl FnMut(usize, f32) -> Vec<f32>,
 ) -> Vec<f32> {
+    let mut measurements = HashMap::new();
+    optimize_fitted_widths_with_cache(
+        baseline,
+        desired,
+        minimums,
+        row_count,
+        &mut measurements,
+        &mut measure_column,
+    )
+}
+
+fn optimize_fitted_widths_with_cache(
+    baseline: &[f32],
+    desired: &[f32],
+    minimums: &[f32],
+    row_count: usize,
+    measurements: &mut HashMap<(usize, u32), Vec<f32>>,
+    measure_column: &mut impl FnMut(usize, f32) -> Vec<f32>,
+) -> Vec<f32> {
     const WIDTH_STEP: f32 = 8.0;
     const MAX_PASSES: usize = 8;
     const MAX_PAIR_STEPS: usize = 8;
-    const MAX_MEASURED_CELLS: usize = 4_096;
 
     if baseline.len() < 2
         || baseline.len() != desired.len()
@@ -817,17 +835,19 @@ fn optimize_fitted_widths(
     }
 
     let mut widths = baseline.to_vec();
-    let mut measurements = HashMap::new();
-    let measurement_limit = MAX_MEASURED_CELLS / row_count;
-    if measurement_limit < baseline.len() {
+    let measurement_budget = MAX_TABLE_MEASURED_CELLS / row_count;
+    if measurement_budget < baseline.len() {
         return baseline.to_vec();
     }
+    // Phase 2 reuses measurements shared with earlier width candidates, but
+    // every candidate receives the same bounded allowance for new layouts.
+    let measurement_limit = measurements.len().saturating_add(measurement_budget);
     let mut score = table_height_score(
         &widths,
         row_count,
-        &mut measurements,
+        measurements,
         measurement_limit,
-        &mut measure_column,
+        measure_column,
     )
     .expect("the baseline fits the checked measurement budget");
     let mut best_widths = widths.clone();
@@ -837,7 +857,7 @@ fn optimize_fitted_widths(
         // Widening a column that is below every current row maximum cannot
         // reduce table height. Recompute after each accepted move so a column
         // that becomes the new maximum remains eligible on the next pass.
-        let relevant_receivers = columns_at_row_max(&widths, row_count, &measurements);
+        let relevant_receivers = columns_at_row_max(&widths, row_count, measurements);
         let mut best: Option<(TableHeightScore, usize, usize, f32)> = None;
         for donor in 0..widths.len() {
             let donor_room = widths[donor] - minimums[donor];
@@ -868,9 +888,9 @@ fn optimize_fitted_widths(
                     let Some(candidate) = table_height_score(
                         &candidate_widths,
                         row_count,
-                        &mut measurements,
+                        measurements,
                         measurement_limit,
-                        &mut measure_column,
+                        measure_column,
                     ) else {
                         return best_widths;
                     };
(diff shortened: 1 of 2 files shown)


aydiler:
Reviewed on a rebase onto current `main` (`7c76e01`), not on your base — `main` moved by one commit since you branched, and #181 only touched `src/main.rs`, so the rebase is clean and the control is current rather than your base. Baseline after rebase: **145 passed, 1 ignored, 0 failed** across the four renderer test binaries, with the CI feature set.

## The red runs you were asked for

A regression that has never been observed red proves nothing, so I mutated the implementation three ways and recorded which tests notice. Every mutation was applied to the production code only, never to the tests.

| mutation | lib | `tests/wrapping.rs` |
|---|---|---|
| **`TABLE_OVERFLOW_FRACTION` 0.30 → 0.25** | 3 red: `…obey_all_caps`, `…reproduces_a_bounded_overflow_knee`, `…scores_every_candidate_through_a_dip_and_recovery` | 1 red: `dense_markdown_table_uses_bounded_scrollable_width_for_wrapping` |
| **early exit at the first non-improvement** | 3 red: `…only_engages_below_the_header_floor_total`, `…reproduces_a_bounded_overflow_knee`, `…scores_every_candidate_through_a_dip_and_recovery` | green |
| **Phase 2 never engaged** (activation forced to the fitting branch) | 4 red, the three above plus `…cache_does_not_remeasure_a_stable_layout` | **2 red: both `dense_*` integration tests** |

The last row is the one that mattered most to me, and it is the reason I ran the integration binary separately: `cargo test` stops at the first failing target, so the first two mutations never reached `wrapping.rs` at all. With Phase 2 disabled both integration fixtures fail on their **width** assertions (`wrapping.rs:219` and `:236`) — which is the proof that they sit inside the dense regime and are measuring your change, rather than passing for a reason unrelated to it. That is the trap #157's first three fixtures fell into from the other direction, and your fixtures are clear of it.

The 25 % row also lands exactly where the arithmetic said it would: `dense_table_reproduces_a_bounded_overflow_knee` is red because +160 becomes unreachable on a 598 px floor total. The policy can be re-derived from the tree now, which was the other half of the ask.

## Policy conformance

All six points check out. Four are covered by your tests; two I verified by reading, because no test asserts them:

- **viewport unchanged** — both call sites still carry `.max_width(table_bound)` and `table_scope_rect.max.x = min.x + table_bound`; only the column widths grow, so the extra width is scrollable overflow. Unchanged by the diff.
- **height reservation follows the new widths** — `reserved_row_heights` measures against `initial_widths`, so the reservation and the paint agree. Worth stating explicitly because a divergence there is what #116 turned out to be.

Two smaller notes, neither a problem:

- The horizontal scrollbar is not new in this regime. Activation requires `minimum_total > visible_column_budget`, and `fit_column_widths` clamps to the floors when `available <= minimum_total`, so a dense table already overflowed its budget before Phase 2. Phase 2 only makes the existing overflow larger, bounded by the cap.
- `layout_revision` only advances when an image size is first observed or changes, so the 6× Phase 1 cost is paid per image-load settle, not per frame.

## One finding

`table_layout_key` collapses every dense budget onto one key by hashing `visible_column_budget.max(minimum_total)`. The reuse is right and `…cache_does_not_remeasure_a_stable_layout` pins it — but `.max()` and `.round()` together also collapse the **fitting** side of the crossover into the same bucket for budgets in `[minimum_total, minimum_total + 0.5)`, and those budgets take the other branch. Two probes I added temporarily and then removed:

```
tmp_probe_crossover_key_collision  ... FAILED
  assertion `left != right` failed
   left: 2516161417735543338
  right: 2516161417735543338      # budget 199.6 (dense) vs 200.4 (fitting)

tmp_probe_crossover_reuses_dense_widths ... FAILED
  a fitting budget inherited the dense overflow:
  dense=[108.0, 152.0] sum=260   fitting=[108.0, 152.0] sum=260
```

With `minimums = [100, 100]` the 200.4 px budget gets the 260 px dense layout — a 60 px overflow (the 30 % term on a 200 px floor total) on a budget it should have fitted exactly. Dragging the divider through that half-pixel band shows it for as long as the width stays inside; one pixel further and the key changes and it self-corrects. So: transient, low severity, but it is a branch decision that the key does not encode.

Hashing the decision itself keeps the intended reuse and removes the boundary:

```rust
let dense = minimum_total > visible_column_budget + 0.01;
dense.hash(&mut hasher);
if dense { minimum_total } else { visible_column_budget }
    .round()
    .to_bits()
    .hash(&mut hasher);
```

Your existing cache test extends to it directly with the 199.6/200.4 pair.

## Also worth a comment in the code

The `else { break; }` on a `None` from the per-candidate `table_height_score` reads like an early exit, which is the one thing the policy forbids. It is not one: the limit is recomputed as `measurements.len() + MAX_TABLE_MEASURED_CELLS / row_count` **after** Phase 1, so a score call can only refuse when that budget is smaller than the column count — which is a property of the table, identical for every candidate, so the break fires on candidate 0 or never. Your `…fails_closed_when_a_full_score_exceeds_the_budget` covers the candidate-0 case. Please say that in a comment, otherwise the next reader either "fixes" it into a real early exit or removes the guard.

## Other checks

- **Test inventory by name**, same command both trees: nothing on `main` is missing from the branch, exactly your ten added, 136 → 146. Counts alone would have been misleading here since the binaries differ.
- **clippy**, CI's invocation with `--all-targets`: **23 warnings before, 23 after**. The only movement is `cached_height_aware_widths` going from 8 to 9 arguments; `table_layout_key` stays at 8 because `visible_column_budget` replaces `table_bound` rather than adding to it.
- **No fork version bump needed.** `main` already carries 0.28.1 (#180) and crates.io is still at 0.28.0, so this rides along on the next release.

Holding the merge only on the key-collision line — the rest is done and I have no further asks. If you would rather ship it as-is I will merge and fix that myself, just say so.

RichardCao:
Thanks for the detailed review and mutation checks. I fixed the crossover collision in `82137bd` and rebased the PR onto current `main` (`d5e3229`).

The optimizer gate and cache key now share one `table_floors_overflow` predicate. The key hashes the dense/fitting branch before hashing the normalized budget, so 199.6 px (dense) and 200.4 px (fitting) remain distinct even though both numeric buckets round to 200.

I extended the cache regression with that exact pair. It first stores the 260 px dense result, then verifies that the 200.4 px fitting layout is remeasured and sums to 200.4 rather than inheriting the dense widths. I also added a focused key test. As a red-before-green check, temporarily removing the branch hash reproduced the identical-key failure; restoring it made both tests pass.

I also documented why `table_height_score == None` is a table-wide fail-closed budget condition (candidate zero or never), rather than a curve-based early exit.

Validation after the rebase:

- root: 67 passed, 1 ignored
- renderer with math: 93 passed
- wrapping with math: 30 passed
- backend with math: 22 passed plus 1 doc-test
- independent high-depth review: no blockers

aydiler:
Verified and merging. I re-ran everything rather than reading the summary, because a fix in the same function can move the guards that cover it.

**The finding is closed, checked with my own probes rather than your tests.** The two throwaway probes that demonstrated the collision yesterday — identical keys `2516161417735543338`, and a 200.4 px budget inheriting the 260 px dense layout — both pass on `82137bd`:

```
tmp_probe_crossover_key_collision       ... ok
tmp_probe_crossover_reuses_dense_widths ... ok
```

Hashing `dense` before the normalized budget is the right shape, and folding the gate and the key onto one `table_floors_overflow` predicate is better than what I suggested: the two can no longer drift apart, which was the underlying reason the boundary went wrong in the first place.

**The mutation matrix still bites after the rebase** — this is the part that needed redoing, since your change touches the function the guards cover:

| mutation | lib | `--test wrapping` |
|---|---|---|
| reference | 146 passed, 1 ignored | 30 passed |
| cap 0.30 → 0.25 | 4 red | — |
| early exit at first non-improvement | 4 red | — |
| Phase 2 never engaged | 4 red | **2 red — both `dense_*` fixtures** |
| **branch hash removed** | **2 red** | — |

The last row reproduces your own red-before-green independently: removing `dense.hash(&mut hasher)` fails both `dense_overflow_cache_does_not_remeasure_a_stable_layout` and the new `table_layout_key_separates_rounded_dense_and_fitting_budgets`. The matrix is also *more* sensitive than before — the extended cache regression now joins the first three rows, where previously it only reacted to Phase 2 being disabled entirely.

**Both integration fixtures still sit in the dense regime.** Worth restating because it is the check that is easy to lose in a rebase: with Phase 2 disabled they both fail on their width assertions, so they are measuring your change rather than passing for an unrelated reason.

**On the `None` guard:** the comment lands exactly where I wanted it, and states the reasoning I had derived separately — identical per-candidate budget, so candidate zero or never. That is the sentence that will stop someone "optimizing" it into a real early exit.

Other checks: CI five of five on your head; test inventory compared **by name** against `main` — nothing lost, eleven added, 136 → 147 (counts alone would have been useless here, since the four binaries report separately); clippy **23 warnings before and after** on a forced re-check, 0 errors; and no fork version bump needed, since `main` already carries 0.28.1 while crates.io is at 0.28.0.

Phase 2 of #118 is done. Thanks for the arithmetic on the 30 % cap back on #118 — that was the part I would not have caught, and it is what made the knee fixture able to express its own decision.

PR #189, opened 2026-09-08
The renderer suite no longer needs a feature list

aydiler:
Docs only, one file — and one that is **auto-loaded into every session**, so a stale instruction in it costs every future one.

`.claude/rules/release-workflow.md` still prescribed the long `RENDERER_FEATURES` invocation and told the reader that a plain root `cargo test` "covers only half the code, silently". Both statements are out of date:

- **The feature list is no longer needed.** Since #185 the renderer's own manifest runs with no flags: `cargo test --manifest-path crates/egui_commonmark/Cargo.toml` → **148 passed**.
- **"Half" was never the figure.** Measured on this commit: the root command runs **65**, the renderer's manifest **148**, and the featured configuration CI repeats **135, 1 ignored**. All three re-measured here rather than carried over from the issue thread.

The featured form stays in the file, because it is what the application actually uses and what CI runs — the plain form is default features and is the one for local iteration.

**Two things added that a reader needs and the file did not say:**

- a `cannot find macro commonmark in this scope` failure means a checkout from *before* #185, not a broken tree. That misreading is exactly what #122 recorded as the cost of the layout, and it deserves a line where someone hitting it will look.
- compare test **inventories**, not counts, when checking a branch lost nothing. Counts differ per configuration and per binary, so reading two totals invents regressions — and it hid a real one this session: a duplicated `#[test]` in my own #188 patch silently took the attribute off an existing test, which the count did not reveal and the by-name comparison did.

### DIFF
--- a/.claude/rules/release-workflow.md
+++ b/.claude/rules/release-workflow.md
@@ -73,16 +73,38 @@ Pushing the tag runs `.github/workflows/release.yml`, which publishes to
 
 ## Verification worth doing before tagging
 
-- `cargo test` at the root, and the renderer crates through their own manifest:
+- `cargo test` at the root, and the renderer crates through their own manifest.
+  **Two commands, because they are two workspaces** — the root command has no
+  reason to descend into the renderer, so on its own it leaves the whole
+  renderer untested and says nothing about it:
+
+  ```bash
+  cargo test --locked                                          # the app: 65
+  cargo test --manifest-path crates/egui_commonmark/Cargo.toml  # the renderer: 148
+  ```
+
+  The second command needed a feature list until #185; it no longer does.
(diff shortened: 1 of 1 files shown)


aydiler:
Force-pushed once: the branch was first pushed while still based on `be4f6f9`, one commit before #188 merged, so the PR's diff included `src/main.rs | 118 -------` — it would have reverted #188's characterization table. Rebased onto current `main` and re-pushed; the diff is now the one file, 26+/4-.

Recording it rather than quietly fixing it, because it is the same shape as the trap in `docs/LESSONS.md`: a rebase that git reports as clean can still be wrong, and `git diff --stat` against *current* main is the check that catches it.

